PR #4, opened 2015-06-07
Mkversion for windows/linux

foxbird:
Added mkversion for windows.
Compiles and runs on both windows and linux.
Needs to be integrated into the linux makefile.
Reads from version.tpl to build the version.c file.
Changed windows build to use C compiler instead of CPP
Reworked the makefile.win and moved it out of the win32 directory. Now build with nmake /f makefile.win
Fixed errors/warnings with main code and windows C compiler.

To add mkversion to the linux build, simply compile mkversion.c and sha1.c together into mkversion. Make sure to run it in the 'src' directory. It will build and create the version.c file. Then continue with the build. It may be worthwhile to make version.c depend on mkversion (the executable), version.tpl, and all of the .o files. This way it rebuilds version.c every time anything else is changed. See makefile.win for an example.


fbmuck/src/version.tpl
```diff
@@ -0,0 +1,112 @@
+/*
+ * Copyright (c) 1991-2009 by Fuzzball Software
```
kyanha: Perhaps the copyright line should be updated "1991-2015"?

foxbird: Just copied this from the existing version.sh/c file. I can update this in a later pull request, or someone else on the project can if they so choose.

foxbird:
Other things that need doing after this patch:
* Change copyright in version.tpl
* Change stricmp to strcasecmp for compiling on mac's
* Clean up assignments as booleans if (!(blah = fun())) for pedantic compilers and settings
* Move the #ifdef WIN32 macros in the mkversion.c into the win32.h file if not already
* Add (sha1) or (git) to end of the hash displayed in do_version so it's obvious which hash is shown
* Add *.h files to readdir/FindFille calls to version those as well
* Add fb-help and fb-resolver into Makefile

I'll be happy to do these things, once this outstanding pull request gets these changes in.

foxbird:
Closing this pull request to refactor the above issues as well as requested changes to use do_hash function instead.

PR #7, opened 2015-11-14
Updated fuzzball.cnf with more secure defaults

foxbird:
Updated the fuzzball.cnf file with some better defaults. 4096 bits and sha256 over sha1. These will provide more security for people making their first certs.

### DIFF
--- a/fbmuck/fuzzball.cnf
+++ b/fbmuck/fuzzball.cnf
@@ -1,5 +1,6 @@
 [ req ]
-default_bits		= 2048
+default_bits		= 4096
+default_md		= sha512
 distinguished_name	= req_distinguished_name
 #attributes		= req_attributes
 #x509_extensions	= # The extentions to add to the self signed cert
--- a/fbmuck/fuzzball.cnf
+++ b/fbmuck/fuzzball.cnf
@@ -1,6 +1,6 @@
 [ req ]
 default_bits		= 4096
-default_md		= sha512
+default_md		= sha256
 distinguished_name	= req_distinguished_name
 #attributes		= req_attributes
 #x509_extensions	= # The extentions to add to the self signed cert


woggle:
You should specify sha256 or sha384 instead of sha512. Client support for SHA-512 certs isn't that great (e.g. [iOS 9's TLS library](https://www.ssllabs.com/ssltest/viewClient.html?name=Apple%20ATS&version=9&platform=iOS%209) and [Mozilla's TLS library until this April](https://bugzilla.mozilla.org/show_bug.cgi?id=1155922) didn't support it, but supported SHA-384 and SHA-256), and it's generally believed that 4096-bit RSA is way more likely to be a problem than SHA-384.

foxbird:
I admit that alternatives such as sha256 and a 2048 bit key are likely "good enough" for today's standards as well. This was an attempt at being somewhat forward thinking in the defaults. If folks are attempting to compile/run/connect on older platforms that don't support more modern options, the defaults are easy enough to change. In fact, it should likely be encouraged that they change these to suit their clients. In the event that they aren't, erring on the side of more secure would be worthwhile.

That said, I'm happy to rescind this pull request and submit one for sha256 with a 2048 key as the default instead, if that's more acceptable to people. At the very least, moving away from sha1 should be a must.

woggle:
We definitely need to remove SHA-1, yes. Larger RSA keys I'm indifferent about.

I just want to make sure that the defaults we give have near universal client support, which RSA (any size) and SHA-256 or SHA-384 will. (You need to go back quite far to find clients that don't support those, and such clients will notice that a lot of the Internet doesn't work for them already. This is not the case with SHA-512, for better or worse.)

foxbird:
I totally understand. I'll let you guys decide on this since it's a bigger change than just a windows fix. Whoever the person with decision power is, just let me know and I'll update the branch with the suggested changes. I'm happy so long as the default is stronger than it is now.

foxbird:
Updated to sha256 with 4096 bits, as confirmed by wyld and above comments.

PR #10, opened 2015-11-14
Win32 compile fixes

foxbird:
Here's an explanation of the changes:

* Changed config.h to move the WIN32 stuff to the bottom. There were includes for header files after the win32.h include. So it would redefine some of the POSIX functions which would generate compile issues. This might have just been Visual Studio 2015 being a pain, but since that's released now I moved to it.
* Updated the windows makefile to include the timestamp file that was added, as well as updating the debugging options. Which I needed for some of these other changes.
* Changed the mkversion to include win32.h as the defaults no longer include the needed windows header files. 
* Changed win32.h to check of certain functions are defined before trying to define them. This avoids issues with popen and _popen for example.
* Changed some %ld's to %lld's since time_t is now a 64 bit value.
* And lastly, there as a smattering of commits to chance some 'long' values to 'time_t'. Long should NOT be used to do math on time_t types. The big reason is that the sizeof on long can vary from system to system. Windows, for example, leaves long as 4 bytes, but linux defines it as 8. So, linux would happily let you work with both, but windows would not. This became a problem in the time_format_x functions which took a 'long' and then cast it's pointer into a time_t. This is obviously wrong, and windows gmtime returned NULL (likely due to the 8 by dereferencing into a 4 byte value on the stack). And since the return value of gmtime isn't checked, segfault. So, to clean it up I looked at a number of places where time_t values were worked on and stored in longs. There were a few parts dealing with the mpi/muf profiling that also worked with doubles (converting to %'s), but they seemed to subtract two time_t's from each other, thus resulting in a number I thought would be small enough to fit into a long. Someone might want to look into that and perhaps rework those parts to be more friendly with time_t calculations.

### DIFF
--- a/fbmuck/include/config.h
+++ b/fbmuck/include/config.h
@@ -201,18 +201,6 @@
 /************************************************************************/
 
 /*
- * Windows compile environment.
- */
-#ifdef WIN32
-#undef SPAWN_HOST_RESOLVER
-#define NO_MEMORY_COMMAND
-#define NO_USAGE_COMMAND
-#define NOCOREDUMP
-#define SCARY_MUF_PRIMS /* Most Windows users don't have their own compiler */
-#include "win32.h"
-#endif
-
-/*
  * Very general defines 
  */
 #define TRUE  1
@@ -366,6 +354,18 @@
 
 #ifndef SYS_TYPE
 # define SYS_TYPE "UNKNOWN"
+#endif
+
+/*
+ * Windows compile environment.
+ */
+#ifdef WIN32
+#undef SPAWN_HOST_RESOLVER
+#define NO_MEMORY_COMMAND
+#define NO_USAGE_COMMAND
+#define NOCOREDUMP
+#define SCARY_MUF_PRIMS /* Most Windows users don't have their own compiler */
+#include "win32.h"
 #endif
 
 #endif /* _CONFIG_H */
--- a/fbmuck/include/db.h
+++ b/fbmuck/include/db.h
@@ -543,7 +543,7 @@ struct frame {
 	short skip_declare;         /* tells interp to skip next scoped var decl */
 	short wantsblanks;          /* specifies program will accept blank READs */
 	dbref trig;					/* triggering object */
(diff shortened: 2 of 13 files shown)


foxbird:
Unfortunately it doesn't seem that the printf family of functions have an argument for taking time_t values. If they did, it would solve the problem entirely. Unfortunately, just casting time_t to a 'long long' poses the same sorts of problems for future development that just assuming it's a 'long' did. The good news is that we can pretend the problem doesn't exist until some really long time in the future :) I just don't prefer to 'pass the buck' as it were.

I think a proper solution would be to test the sizeof time_t and select either %ld or %lld as appropriate. If I recall, linux types have autoconf which might be able to do that, and substitute where appropriate, or at least provide an #ifdef to test it with. I'm not that adept with autoconf though, so someone else may want to look at that.

Also, how many systems out there that we are targeting aren't using a 64 bit time_t yet anyway? 2038 isn't tomorrow or anything, but linux has been sporting updated time_t for a while now. I'm just curious what the targets are so that I can research them before making changes. With the rather liberal treatment of time_t as long, as well as the windows functions not quite the same as linux, I'm sure there will be more reason for changes. I'd rather try and get them right first go round to save you guys the headache of spot checking it all.

woggle:
I think the explicit cast solution is easiest, and the only risk seems to be if time_t become a larger integer than a long long. We should not select the macro to use based on sizeof(time_t) because systems don't need to pass different types of the same width the same way through varargs.

64-bit Unix-likes usually have time_t be an alias for long and have long be a 64-bit type. Thus there isn't actually a 2k38 problem there. Usually long long is also 64-bit, so this means %lld and %ld are silently the same thing in *printf(), but it's not actually guaranteed that they're passed the same way or that long long won't become a 128-bit type on some future architecture.

On 32-bit Unix systems, there's presumably going to have to be an ABI change which AFAIK hasn't happened yet on any well-known platform.

foxbird:
There yah go. Updated for that issue. Let me know if there are others you encounter.

PR #13, opened 2016-02-01
Add command to reload SSL configuration

woggle:
This pull request includes #12.

Add a Wizard-only command `@reconfiguressl` to reinitialize the SSL_CTX, including reloading certificate files. The new configuration will be effective for all subsequent connections. Notably, this allows certificates to be renewed without restarting a MUCK. It also allows changes to SSL @tunes to be effective without restarting.

In the event that reconfiguring SSL fails, for example because a new certificate file is bad, the old configuration remains active, and a vague error is given to the Wizard running the command and a less vague error appears in the log file.

### DIFF
--- a/game/data/help.txt
+++ b/game/data/help.txt
@@ -32,18 +32,18 @@ ALPHA|ALPHABETICAL|COMMANDS
 You can get more help on the following topics:
 
 Symbols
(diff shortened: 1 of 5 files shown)


woggle:
Updated to make this merge cleanly; fixed problem with compilation with IPv6 disabled.

PR #18, opened 2016-03-07
Amend @tune help; not entirely wizard-only

digitalcircuit:
Change help to indicate that some, not all, @tune parameters are wizard-only.

Servers may override @tune for non-wizard players, but in that case they'd need to change the help file anyways.  When releasing the next version, the changelog should note to update help.

(*Pardon the mistype and recent Windows build break; I'll try to test more options with future changes*)

### DIFF
--- a/game/data/help.txt
+++ b/game/data/help.txt
@@ -1918,8 +1918,6 @@ typed in full.
 @TUNE %<param>
 @TUNE info <param>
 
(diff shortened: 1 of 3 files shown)


digitalcircuit:
After some email discussion, it seems I misread.  Pardon the noise!  I'll close this for now, but it can be reopened in the future if needed.

Issue #40, opened 2016-12-13
Futex wait causing MUCK-wide lockup


sypherspace:
We're experiencing an issue, albeit infrequently, where the MUCK will completely stop responding to input. Attempting to trace the process yields the following:

```
# strace -p 22940
strace: Process 22940 attached
futex(0x7f6882b3bac0, FUTEX_WAIT_PRIVATE, 2, NULL
```

It hangs at this point until strace is interrupted. We've tried launching the MUCK with fbmuck compiled in debug mode, but it fails to read the database when compiled in such fashion. In the mean time, I've pulled down the latest code as of this date/time in the hopes that a committed change has resolved the issue.

wyld-sw:
Could you give us a little more information about your environment? Which Linux distribution, compile options (from @version), etc.

I read there was a linux bug with futex(), but that may be a red herring.

kyanha:
Version: Muck2.2fb7.00a2(2) Compiled on: Sun Dec 18 2016 at 14:18:39 UTC
Options: DETACH GODPRIV IPV6 MCP RESOLVER SSL

Ubuntu 16.04, last patches approximately at the beginning of November 2016.

sypherspace:
Regarding the compile time, I accidentally triggered a rebuild when looking through the config.h header file. We haven't changed the code since the issue was filed. The most recent lockup was approximately 12 hours ago, 12/17/2016, 11:30 PM ET.

woggle:
I'm curious about being unable to load the database with the MUCK compiled in debug mode. Do we have any idea what's causing that? Can we at least get it working compiled with -g -O2 or similar so you can attach a debugger and possibly get a half-useful stacktrace the next time this happens?

woggle:
One possibility that comes to mind is that sig_reap() might do things which are not async-signal-safe, like calling vfprintf via log_status, which might cause a deadlock if the MUCK gets SIGCHLD while writing a log message (or malloc'ing if vfprintf mallocs, etc.)

A likely symptom of this would be the freeze occuring shortly after a forked database dump or a resolver process exit.

sypherspace:
Actually, it does seem to happen in conjunction with a database dump to disk, based on timing.

kyanha:
There's also a perhaps-related issue with resolver not dying properly?  I haven't dug in to figure out where or why that's happening -- @sypherspace, can you submit a separate bug for the resolver issues?

sypherspace:
[...]

PR #41, opened 2016-12-18
Use pselect() for SIGCHLD, SIGUSR1

woggle:
Our handlers for signals do async-signal-unsafe things, which might be the root cause of MUCK freezes on futex() calls reported in issue #40. One solution to this is to ensure the signal handler only runs at a safe time, such as when we would call select(), rather than processing it immediately. This patch tries to implement that solution for SIGUSR1 and SIGCHLD on systems that support pselect().

SIGUSR1 and SIGCHLD are masked out of the process except during a call to pselect() (which replaces calls to select()), so the signal handler should only run during the call to select(). I only do this for SIGUSR1 and SIGCHLD because these are signals we expect the MUCK to continue running after; for signals like SIGSEGV or SIGTERM it seems more important to attempt to do something right away.

### DIFF
--- a/configure
+++ b/configure
@@ -660,7 +660,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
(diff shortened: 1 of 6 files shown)


kyanha:
The rundir should also be where the fbmuck.pid gets placed.  If it is moved, should we warn people that it's going to move so they can fix whatever external scripts they've got going on (FurryMUCK's powerfail script is the main one that I'm aware of, even though I don't think it's going to get invoked ever again now that they're on Linode)?

woggle:
I just ran autoreconf -fi, I didn't expect the changes to configure to be substantive. Apparently Debian and Ubuntu has backported this autoconf feature (to be included in autoconf 2.70). I just removed it manually it to avoid confusion (since we don't actually use it).

PR #65, opened 2017-02-07
    Fix looping issues with new waitpid() loop in sig_reap()

kyanha:
It turns out that waitpid() returns 0 if called with WNOHANG and
    there are no errors.  The prior commit only checked for -1, which
    is returned only on an error.
    This commit removes a spurious if(), and makes errno available
    to be logged if waitpid() actually returns -1.

### DIFF
--- a/include/fbsignal.h
+++ b/include/fbsignal.h
@@ -1,6 +1,18 @@
 #ifndef _FBSIGNAL_H
 #define _FBSIGNAL_H
 
+#ifndef WIN32
+#ifdef HAVE_ERRNO_H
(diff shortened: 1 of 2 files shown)


kyanha:
Most of the changed lines are just changing the indent level for them.  I don't know why the display is so bent on showing them as deleted/added.

I needed to add <errno.h> or <sys/errno.h> to get the bare minimum useful diagnostic, and I'm still not sure if there's a way to change the value back into the symbolic name.  If there is, that should be used instead because errno values are different from platform to platform.

PR #113, opened 2017-03-25
check if object is program before trying to compile it from $ifcancall

woggle:
This makes specifying a non-program with $ifcancall a compile error. I'm not sure this is the right thing to do instead of making it just consider the "can call" part false.

Fixing this fixes a segfault when trying to compile `$ifcancall me 0`.


wyld-sw:
Changes committed to take the false branch for non-programs. Also disallows "this" as an argument, since re-compiling the same program into oblivion seems like a bad idea.

PR #143, opened 2017-04-23
prim_diff3: do not trip stack pop depth debugging checks on errors

woggle:


woggle:
#144 is probably a better solution to this.

PR #160, opened 2017-04-29
Limit depth of MUF interpreter nesting for locks

woggle:
Also turns max interpreter nesting limit into an `@tune`.


woggle:
Er, I see that this is redundant with code in TESTLOCK.

I think problems still exist via MPI, but a consistent way to address with the current code would be checks on `fr->level` before do_parse_mesg calls. I'll aim for a different patch.

PR #192, opened 2017-10-20
Added MCP package org-fuzzball-languages.

revarbat:
The MCP org-fuzzball-languages package allows clients to query what extension languages the server supports.

The client sends the org-fuzzball-languages-request message with no args, and the server will send back an org-fuzzball-languages-supported message with a languages list argument, where each line is a language name and a version number separated by a colon.  Currently, the only response is "muf:7.0".


revarbat:
At some point, of course, the list of languages supported should be pulled from a header file somewhere, which can be used by the $language directive as well.

PR #218, opened 2018-06-04
Fix attributes list in mufman.raw (also a question in the extended de…

nyanster:
…scription)

Question: does anything convert the .raw files to the .txt files, or is that only for .raw to .html?  Duplicated text/code is never fun!

### DIFF
--- a/game/data/mufman.raw
+++ b/game/data/mufman.raw
@@ -2523,9 +2523,9 @@ series of comma separated attribute names.
 
 These are the attributes currently recognized:
 ~~code
-    reset    bold   dim       uline    flash   reverse
-    black    red    yellow    green    cyan    blue    magenta    white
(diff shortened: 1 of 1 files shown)


wyld-sw:
My understanding is that 'make help' generates both the html and txt files.

nyanster:
I see.  You're right: I've done that and confirmed it is so.  Should the .txt versions be deleted from github then?  Or, how else do you ensure that the .txt versions don't fall behind the .raw versions?

digitalcircuit:
`make help` does not (yet) work on Windows, unfortunately, otherwise we could have the help files always generated at compile time.

It might be worthwhile fixing that first before removing the generated `.txt` files to avoid leaving Windows builds (including automated AppVeyor builds) out of the picture.

wyld-sw:
Thanks, Shane. Good point!

Issue #206, opened 2018-04-02
Allow for specification of a keepalive command


hyperlumic:
Server-side TCP keepalives can help, but we have a few players that need client-side keepalive capability. I'd like to see a command or a tuneable value that lets us set a command that a player can trigger via client-side timer that generates regular network activity, but does NOT reset their WHO idle time.

nyanster:
Maybe they should actually be active and inputting commands when they're on the MUCK, instead of idling. That'd be better than a keepalive.
This idea promotes destroying immersion on MUCKs, and MUCKs are a form of virtual reality.  Maybe individual MUCKs can implement this, but why do this in FuzzBall itself?  It's a self-defeating idea.

hyperlumic:
Because some poorly-configured consumer-grade routers have obscenely short TCP keepalive values, for one. Some terminate idle sessions after just 60 seconds. Don't know about you, but it can take me a couple minutes to write a pretty in-depth pose. Or maybe I'm working on a MUF program in a text editor, and pasting the code into the MUCK when I want to test. For this and many other reasons, I, as the server administrator, should have the ability to deal with these cases.

Also, the reason I specified the `NOT reset their WHO idle time` criteria was so that this would not cause a user to circumvent the idleboot time value on configured MUCKs.

> Maybe individual MUCKs can implement this, ...

Without the ability to exempt a command from resetting a user's idle time, how do you propose to do so?

nyanster:
> Also, the reason I specified the NOT reset their WHO idle time criteria was so that this would not cause a user to circumvent the idleboot time value on configured MUCKs.

Good point.  My bad.

wyld-sw:
NULL_COMMAND (when recognized) should no longer reset idle time.

hyperlumic:
I just wanted to confirm, first. It still appears to reset the idle time when used.

wyld-sw:
It does in some cases, but not all it seems. Reopening until we can figure this out. Thanks for the report.

wyld-sw:
Is this fixed? Seems like the commit by @evenfl0wz addresses it, but I'm getting conflicting results in my environment.

PointsDragon:
[...]

PR #240, opened 2018-12-01
Fixed NULL_COMMAND resetting idle timer

syversed:
This should address  #206.

The idle timer should no longer reset if the input command is the same as NULL_COMMAND, regardless of what NULL_COMMAND is set to.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -1844,10 +1844,6 @@ process_input(struct descriptor_data *d)
 		|| strncasecmp(d->raw_input, NULL_COMMAND, got)
 		&& !(got-2 == strlen(NULL_COMMAND))) {
 		d->last_time = time(NULL);
-		log_status("COMMAND: %s", d->raw_input);
-		log_status("GOT BYTES: %i", got);
-		log_status("NULL LEN: %i", strlen(NULL_COMMAND));
-		log_status("MATCHES? %i", strncasecmp(d->raw_input, NULL_COMMAND, got-2));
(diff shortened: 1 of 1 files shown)


wyld-sw:
I'm not sure this completely solves it. I issued several `@@`s in a row and at least one of them reset the idle timer. Does that happen for you?

evenfl0wz:
I've issued several ```@@``` in a row and I don't seem to be able to reproduce this behavior.

PointsDragon:
Not fixed - if NULL_COMMAND is `@@` (two bytes), but you actually need to send `@@\n` (three bytes) to process it - which means 'got - 2' as your comparison is likely not what you want.  Furthermore, if there was a large timeslice, the amount of pending text from the socket read could be much larger, or conversely, if there's a a fast cycle and you're on an unbuffered connection, you may literally be sending a single byte per loop - in which case got - 2 is an overflow/dangerous value.  If we're on a system where malloc doesn't zero init memory, and this is your first pass through on a single byte - this code could actually core dump in a non-obvious manner.

Since you're expected the head of `d->raw_input` to have you N prior bytes to be the NULL_COMMAND here, the proper way to check would be to do your check of raw_input against the sizeof NULL_COMMAND.  Also, since NULL_COMMAND is a constant, why are we doing a strlen of it versus the socket read return?  Was the strlen actually meant to be on raw_input?  It feels that the prior code was 'more correct' than the patch here.

PR #241, opened 2018-12-03
Fix typos in MPI documentation

nyanster:
I only fixed them in the .raw.  The changes will show up in the .txt, .html, and other files whenever someone does the usual steps to build fuzzball.


wyld-sw:
Windows users don't have access to prochelp, if I recall correctly, so please commit the other files as well - until there's a cross-platform solution. Thanks!

PR #248, opened 2018-12-12
Update autostuff

nyanster:
I couldn't figure out how to get autoconf to find the autoconf-archive macros, so I couldn't regenerate the configure script.
Wyld, please help by generating the configure script!  And also by marking everything in the auto folder executable. I don't think anything is using the scripts in the auto folder though... They might be removable.


wyld-sw:
Seems like install-sh is required by the current config script. My interpretation is that the others are outdated or unnecessary at this time. I'm removing those, but I can revert if it turns out we need them.

PR #252, opened 2018-12-13
Update .travis.yml

nyanster:
I've seen some bugs more obviously show up during make install.  I remember in some of my builds, seeing prochelp segmentation fault during make install (which includes make help) let me know something was wrong with a diff I wrote.
Including the very quick make install lets such a chance be possible, to see prochelp segfault or run OK.
Also, switch to the newest version of Ubuntu that Travis has available.
btw is there a difference between make and make all for fuzzball?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 sudo: false
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


nyanster:
Let me delete this and try again, I need to change the configure --prefix for make install to work with travis.

wyld-sw:
I usually use PWR, for better or worse, for the prefix during testing. 

make all specifically recurses into the arc directory, but it might be functionally the same as make (I’ve never specifically used make all). If it is we can probably just use make instead of make all.

PR #253, opened 2018-12-13
Update .travis.yml attempt 2

nyanster:
I've seen some bugs more obviously show up during make install.  I remember in some of my builds, seeing prochelp segmentation fault during make install (which includes make help) let me know something was wrong with a diff I wrote.
Including the very quick make install lets such a chance be possible, to see prochelp segfault or run OK.
Also, switch to the newest version of Ubuntu that Travis has available.
btw is there a difference between make and make all for fuzzball?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 sudo: false
 addons:
   apt:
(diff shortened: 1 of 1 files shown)


digitalcircuit:
Thanks for updating this!

And there's no difference between `make` and `make all` for Fuzzball - `make` [builds the first target by default](https://stackoverflow.com/questions/6856263/what-is-the-difference-between-make-and-make-all), and `all:` is the first target configured.

PR #263, opened 2018-12-20
Change matrix order

nyanster:
This way, the first things that get built are both the nothing and all settings of the matrix, thus providing the most different builds first, with all the other variations after.
Waiting for the nothing/empty option, last, sometimes took a few minutes. :3

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,7 @@ compiler:
 env:
   global:
   matrix:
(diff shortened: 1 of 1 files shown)


nyanster:
Ah okay.

wyld-sw:
DIdn't mean to close without committing. If you re-upload, I'll do it.

nyanster:
Oh, okay!  I was worried there was something more to the order than I saw. >w<
Re-did it. :D

PR #265, opened 2018-12-20
Enable Cirrus CI to get FreeBSD builds

nyanster:
I only have a matrix size of two builds here since Travis already does many.
The FLAG_NONE btw is to get the empty matrix to work.
Wyld, after you merge this, while logged in:
https://github.com/marketplace/cirrus-ci
Enable/install this, the Public Repositories plan.
After that's all done, click 'log in' here:
https://cirrus-ci.com/
And I believe that's everything to enable Cirrus.


nyanster:
You'll need to be logged in as fuzzball-muck (or the owner of it?) to do that btw.  Adding this in in case Wyld and fuzzball-muck are separate accounts.
Not entirely sure.

PR #267, opened 2018-12-21
Enable time.h all the time

nyanster:
See if this unbreaks prev commit. :D


nyanster:
durp, I'll try again after.

PR #270, opened 2018-12-21
Windows doesn't like sys/time.h

nyanster:


nyanster:
The failing cirrus check is just an oddly terminated one.  I feel this diff works. (it even works on the other of the two cirrus checks!)

nyanster:
If I remember correctly when I was using it, fuzzball-muck could login to cirrus-ci.com and manually reflag that check to be re-run, if desired.

wyld-sw:
Good idea. I'd like to move it slightly to be with the other non-WIN32 headers. Commit incoming.

PR #275, opened 2018-12-24
Always include string.h and strings.h

nyanster:
Also eliminate the antique memory.h
future directions: remove STDC_HEADERS, HAVE_STRING_H, HAVE_MEMORY_H


nyanster:
Don't merge.  Windows doesn't have strings.h nor does it have strcasecmp, working on a solution

PR #278, opened 2018-12-24
CI check 2 (nomerge)

ghost:


digitalcircuit:
Small note, you should be able to re-use a single pull request for CI checks.  Force-pushing generally causes a rebuild, as does closing and re-opening.  And if I'm misunderstanding,

*And thanks for all the work!  I've had a busy past month+, but I'll try to catch up soon* :slightly_smiling_face:

PR #284, opened 2019-01-06
Better adhere to Unix hier

nyanster:
Move restart command to sbin.  Renamed it so it won't conflict with other programs on a Unix system that might have restart scripts.  Move documentation files to share/doc/fbmuck rather than share/fbmuck
FUTURE DIRECTIONS:
move fbmuck and fb-resolver to sbin
have help files not originate from game/data/, but maybe from docs/
figure out why GNU autostuff uses "install -c" instead of "install", even for other uses...

### DIFF
--- a/README_LINUX.md
+++ b/README_LINUX.md
@@ -61,8 +61,8 @@ mv "$PREFIX/game/data/basedb.db" "$PREFIX/game/data/std-db.db" # Rename database
 ```
 * Start Fuzzball
 ```sh
-/usr/local/sbin/fbmuck/fb-restart
-# If using a custom prefix, $PREFIX/sbin/fbmuck/fb-restart
+/usr/local/sbin/fb-restart
+# If using a custom prefix, $PREFIX/sbin/fb-restart
 ```
 
 [help-mink]: http://www.rdwarf.com/users/mink/muckman/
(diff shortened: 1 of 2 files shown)


kyanha:
Generally, the docs don't tend to change between different invocations of the server.  That'd typically make them candidates for /usr/share/docs on e.g. a Debian package.  (just a thought.)

nyanster:
> Generally, the docs don't tend to change between different invocations of the server. That'd typically make them candidates for /usr/share/docs on e.g. a Debian package. (just a thought.)

True!  That's exactly what this diff does.

That goal, too, is what I'm slowly working towards.  I eventually want to make Fuzzball into a FreeBSD package and an OpenBSD package, and just more easily maintainable for package maintainers.
Though I wish there were more localdocs, like, for example, something documenting all the commands that come in advancedb that don't come in minimaldb.  Of course most of those commands have their own help options, so this may or may not be useful.  (more useful would probably be having a program list that lists all available programs by the names they can be called by)

Just a potential typo, but
https://manpages.debian.org/stretch/manpages/hier.7.en.html
says that it'd be /usr/share/doc and not /usr/share/docs on Debian GNU/Linux.
The FHS also agrees:
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s11.html
I have Ubuntu on another hard drive but can't easily check the actual file hierarchy at the moment.

aidanPB:
As an Ubuntu user, I can confirm that it's /usr/share/doc.

aidanPB:
As an Ubuntu user, I can confirm that it's `/usr/share/doc`.

PR #294, opened 2019-01-14
Add prop-name-ok? Primitive

### LINKED ISSUE #289: Feature Request: prop-name-ok? primitive
Hey guys;

So there is ext-name-ok?, pname-ok?, and name-ok? (which is deprecated) but no primitive for checking if a string is a valid prop name.  Props can't have : in the name for instance, whereas objects can.

We could either take ext-name-ok? and add a "prop" type, or we could make another primitive called something like prop-name-ok?  I think the latter is slightly more intuitive though the former is probably a little easier.

I would be willing to take a stab at writing the code here, but I'd like the group's input on what the preferred solution is before doing the work (assuming you guys agree with me that this is a necessary feature).

Thanks guys!

tanabi:
This commit resolves #289 

I'm not sure why my previous pull request for removing the propqueue check seems to want to ride along with this PR ... It looks like it was merged into master.  I don't think it'll hurt anything when it comes time to merge this.

HOWEVER ... its not time to merge it :)  When I run a test program, thusly:

``: main
  "test-prop" prop-name-ok? intostr tell
  "test:prop" prop-name-ok? intostr tell
;``

I get a stack underfly on intostr, thusly:

``Debug> Pid 7: #174 1 ("") INIT FUNC: main (0 args)
Debug> Pid 7: #174 2 ("") "test-prop"
Debug> Pid 7: #174 2 ("", "test-prop") PROP-NAME-OK?
Debug> Pid 7: #174 2 ("") INTOSTR
Program Error.  Your program just got the following error.
test.muf(#174), line 2; INTOSTR: Stack underflow.
System stack backtrace:
  0) test.muf(#174) line 2, in main(:
  2:   "test-prop" prop-name-ok? intostr tell
*done*``

I must be missing something.  Also, I don't really understand how the MUF help files work.  When I grep around I see "raw", "html", and "txt" files ... do I just update the "raw" file and the txt/html are generated?


tanabi:
Okay -- also weird.  I tested my setprop changes and it looks like on my test MUCK,
``me @``
does some really weird shit ...

```
Debug> Pid 8: #174 1 ("") INIT FUNC: main (0 args)
Debug> Pid 8: #174 2 ("") V0
Debug> Pid 8: #174 2 ("", V0) @
Debug> Pid 8: #174 2 () "test-setprop"
Debug> Pid 8: #174 2 ("test-setprop") "prop"
Debug> Pid 8: #174 2 ("test-setprop", "prop") SETPROP
Program Error.  Your program just got the following error.
test.muf(#174), line 2; SETPROP: Stack underflow.
System stack backtrace:
  0) test.muf(#174) line 2, in main(:
  2:   me @ "test-setprop" "prop" setprop
*done*
```

Did I break that?! ...

nyanster:
oh!  My bad for making so many separate replies, but check each CI please!  Your diff introduces some new compiler warnings about the code.
-Wincompatible-pointer-types-discards-qualifiers in clang
-Wdiscarded-qualifiers in gcc
warning C4090 in microsoft whatever
Those three are in property.c
AppVeyor also shows a warning in p_props.c

tanabi:
@nyanster I actually didn't intend for this to get merged yet, cause I hadn't finished adding to the documentation :D  So I didn't look at any of the CI stuff.  I will have an updated PR with fixes + the documentation bits momentarily.

PR #298, opened 2019-01-23
Improve page's wording for recipient

nyanster:
This wording for sending a location reduces potential confusion with the 'whereis' command available on many MUCKs, and clarifies what page is doing.
In this pull request is also some corrections to the documentation, which incorrectly documented the behavior of basedb/advancedb page instead of fuzzball's page, for sending messages.


nyanster:
This confusion is real rather than imagined, too. :P
I wrote this diff after someone replied to me accidentally paging her my location.

PR #309, opened 2019-02-09
signal-related cleanup

nyanster:
Got rid of sys/signal.h since we include signal.h all the time.  Got rid of RETSIGVAL since those functions are void, returning nothing.
No functional changes here!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -202,7 +202,7 @@ fi
 dnl
 dnl Header files
 dnl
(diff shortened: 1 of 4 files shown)


nyanster:
TODO: make new configure script based on new configure.ac

nyanster:
Travis might need to be manually told to re-run...  I'll try to have that happen now.

PR #311, opened 2019-02-11
Documentation Project Start

tanabi:
This is my first stab at the documentation project.  Please note a few choices I made that you may want to consider before clicking merge. :)  I know how you love clicking merge, Wyld!!

1. I put a header on each file that has a brief description and a little license note.  I think there's probably a more official header floating around... do we want to have an official header or is this little thing fine?

2. I put doc blocks on BOTH the .h declarations and on the .c definitions.  This is my personal preference because I like the documentation to be close at hand when I work on code, but I  know that it also means updating double the docs if you're doing modifications or adding something new.

3. I used documenter format tags, though I haven't 'tested' it yet to make sure it works.  I want to make sure what I do is 'okay' before doing any more work.

4. You may want to get several opinions on this since we're making the standard going forward here.


wyld-sw:
This is an awesome start, thank you very much!

For the headers, I do like that it refers to a separate file so we can make changes and additions without touching the main text in multiple places. I do know that @nyanster had some ideas about licensing going forward, since there's still FB5-era code in play which may not be compatible with the current license. Two choices I've heard for the code are LGPL and the MIT (Expat) license. @naynset, could you fill in the details on this? 

I'm a fan of having the doc blocks both places as well. Theoretically they shouldn't have to change much, once the associated functions become sane.

This certainly is open to other opinions - any reader is welcome to comment here on a preferred way of managing the license and function docs.

Lastly, I'll resist the urge to merge at this time :)

tanabi:
Updated PR to apply some comments Wyld gave me.

include/props.h
```diff
@@ -54,123 +77,1286 @@ typedef struct plist *PropPtr;
 /* Blessed props evaluate with wizbit MPI perms. */
 #define PROP_BLESSED     0x1000
 
+/* You will never want to change these, or you will make your MUCK
```
nyanster: Aren't these here for future internationalization?  I have no idea, I'm just guessing, so, @wyld-sw could you elaborate?  I know you've done some stuff related to this.

nyanster: They might be there to be mnemonic/useful instead of magic symbols, I guess. >w>

include/props.h
```diff
@@ -54,123 +77,1286 @@ typedef struct plist *PropPtr;
...
+ * set values more complex than string or integer, use set_property_nofetch
+ * instead.
+ *
+ * @internal
+ * @param player Object to set property on
+ * @param pname Property name
+ * @param strval String value - pass NULL if you want an integer value
+ * @param value Integer value - ignored if String value is used.
+ */
+void add_prop_nofetch(dbref player, const char *pname, const char *strval,
```
nyanster: Change from type to pname intentional?

nyanster:
What's with all the propqueue/at-register commits when this issue is about documentation? :P  (I think why it doesn't happen to me has to do with the fact that I delete my forked repository between every pull request I make, in order to start with a fresh fuzzball master.  I don't know if it's the most efficient way to do it, but it does avoid having previous commits show up in pull requests)

I agree with @wyld-sw that having the license mentioned in EVERY file is a bad idea right now, and might always be a bad idea.  The current paragraph is very vague too, anyway.  "Fuzzball MUCK Project" isn't an entity AFAIK so in a legal paragraph I'd use this instead:
Copyright (c) 1989-2019 its authors
and then have an AUTHORS file in fuzzball's .tar.gz (and hence in the repository)
I'm not a legal pro :D  I just follow along and read stuff in big open-source projects.

The explanation on licensing: fb5's/TinyMUD's license was pretty unique, and not really in a good way.  Likely with good intentions, it had many stipulations, like if you make any changes to tell the original authors, etc etc.  It reminded me of licenses that end up being incompatible with the GPL by placing a lot of intentional or unintentional requirements upon the authors.
So, sometime in fb6, the license was changed to GPL.  If the GPL is incompatible with fb5's license, then this change wasn't legal, if any code under the TinyMUD/fb5 license still existed, or perhaps even still exists.  I'd have no idea how to audit the code so thoroughly in a time-efficient way.  But it's very likely at least some lines of code still exist.
But, there are good open-source licenses that are far more friendly to being mixed and matched with other licenses.  This includes the LGPL, and copyfree licenses.  e.g. MIT, 2-clause BSD, ISC, etc.  (also, none of those licenses would need an OpenSSL exception, or any sort of linking exception, which would be a great improvement!)
Now, to change a licence, you have to hold copyright over what you're doing that to.  Since every line of code is copyright its original author, and not to any Fuzzball entity, it means you'd have to ask everyone who submitted code under the GPL for permission in order to change the license of GPL'd code in a totally legal way.  (You wouldn't have to ask people who submitted TinyMUD/fb5-licensed code in order to change GPL'd code to a more compatible license, since their code wouldn't be re-licensed: on that topic, retaining that old license in a historical document along with an explanation of the current situation is a good idea: I've been meaning to do that in my own big documentation project :3 )
https://license.openssl.org/
For example, OpenSSL is trying to contact every contributor to ask for permission to re-license their code.  People who don't respond, or people who say "no thanks!", their code is being re-written purely for copyright reasons.
I personally don't mind bending the rules, but a lot of people in the open-source world do care about following the law, including IP law.  And I've learnt a little legal stuff just by hanging out in this world. :P

This licensing stuff should become its own GitHub issue.  I can write one if you want!

As for this actual pull request :B
I'm not a C pro either...  And I also don't know Fuzzball's code well enough to judge a lot of this.  I began skimming over a lot of the explanation comments since I'll have to put a lot more time into it to truly read and understand them all.  I'm not very good at reading code, especially lots of code at once. uwu  So I'm not the best to ask about this.
There's some inconsistent style with the comments, I've seen two different ones in your pull request:
```
/* this
 * multi-line
 * comment with "this" on the first line
 */
versus
/*
 * this
 * multi-line
 * comment with a blank first line
 * (i prefer this one but that's just a stylistic issue so don't take this so seriously for now uwu)
 */
```
There are also comments with two ** at the top.  But the last few lines in those comments with the at symbols makes it look like a convention I just don't know about yet.  What is it?
Also, I see actual changes in code here when I thought this was just supposed to be documentation for now...  Though, variables are documentation so I just ask to confirm that~  In add_prop_nofetch and add_property the variable "type" changes names to "pname".  This does seem intentional, since your comments explain that it makes far more sense.
And as said above, I think anything related to licensing in this pull request should be removed.

src/property.c
```diff
@@ -1294,122 +2337,250 @@ reflist_find(dbref obj, const char *propname, dbref tofind)
...
+                             */
+                            if (!*temp || *temp == ' ') {
+                                break;
+                            }
+
+                            /* Otherwise, this isn't our number */
+                            pat = NULL;
+                        } else if (*pat != *temp) {
+                            /* Pat and temp are no longer in sync -- this
+                             * isn't our number.
```
nyanster: No */ here.  Intentional?  Seems unintentional...

include/props.h
tanabi: Yeah.  Whomever made the header file did a lot of "copy pasta" and 'type' was used in place of name almost everywhere.  pname is correct in all cases and matches the definition of the function which has the correct name.

src/property.c
tanabi: Yeah that was a mistake, I'll fix it.

tanabi: Reguarding the /** comments

Those are called docblocks and that allows the auto-generation of API documentation.  For instance, doxygen will scan the files, find /** comments, and turn them into API docs.  That's also what all the stuff like @ param is -- doxygen picks those up and handles them special.

I'll remove the license stuff too.

nyanster: > Yeah that was a mistake, I'll fix it.

Okidoki! :D

> Reguarding the /** comments
> 
> Those are called docblocks and that allows the auto-generation of API documentation. For instance, doxygen will scan the files, find /** comments, and turn them into API docs. That's also what all the stuff like @ param is -- doxygen picks those up and handles them special.
> 
> I'll remove the license stuff too.

Thank you for explaining!  That's neat.

tanabi: My next commit will have the doxygen build files, and when I'm done with the documentation project, I'll update Makefile to have a target for building docs (Though it will require you to install doxygen)

tanabi: This should fix all the issues you found, adds doxygen build files, and adds fbstrings as well.  Give a look when you've got a sec.  I eyeballed the tests, looks like no warnings ... fingers crossed, this can be merged.

PR #312, opened 2019-02-13
compiler warning test (CI check, don't merge)

nyanster:
Curious to see what warnings we get when we enable them all on Clang, and many (but absolutely not all) on GCC.  Don't merge this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,6 +13,6 @@ task:
   script:
     - ./configure --prefix="$PWD/fuzzpre" $FLAG_SSL $FLAG_MEMPROF $FLAG_DEBUG
     # Speed up CI builds
-    - printf '/CFLAGS=/s/-g//\ns/-O2//\nw\nq\n' | ed -s src/Makefile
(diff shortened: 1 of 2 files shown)


nyanster:
Linking these so I have these sent to my email...
https://travis-ci.org/fuzzball-muck/fuzzball/builds/492837701
https://cirrus-ci.com/task/5729663247712256
https://cirrus-ci.com/task/5166713294290944
And closing so it doesn't get merged.  You can still comment on this if you see something interesting, though.  The Travis checks are -pedantic -Wall -Wextra for both GCC and Clang, on GNU/Linux.  The Cirrus checks are -pedantic -Weverything (which is many more warnings!  it includes -Wall -Wextra and more) on Clang on FreeBSD, and the logs are long enough you need to download them rather than view them in a browser.

wyld-sw:
Thanks, @nyanster. There were many more -Weverything warnings at one time, so it's getting better! I've also addressed most of the static analyzer issues (https://clang-analyzer.llvm.org) I could find.

wyld-sw:
Also note that some of the more basic errors (unused identifier, etc.) can be red herrings. Configuring with SSL uses additional variables, for example. Though, we may want to put the variable declaration inside the #define in that case...

PR #315, opened 2019-02-15
Let fuzzball build when not compiled with TLS

nyanster:
This diff fixes a build problem, and also provides an informative message to wizards when SIGHUP is received by a MUCK not compiled to use TLS.
Also change order of true and false results for easier readability.  Also changed first wall_status message to specify certificate rather than configuration.  Not intentionally, formatting is changed because the github editor decided to use tabs instead of spaces.  Fine with me.

Relevant previous commits:
https://github.com/fuzzball-muck/fuzzball/commit/55c2bebd65c537551e48a6f2fa40b9131314edf4
https://github.com/fuzzball-muck/fuzzball/commit/3531244046b2e81e5117cfffc01911f6f2dccc09

Results that show Fuzzball building only when compiled with TLS:
https://travis-ci.org/fuzzball-muck/fuzzball/builds/493378579

Compiler error:
https://travis-ci.org/fuzzball-muck/fuzzball/jobs/493378580#L365
https://travis-ci.org/fuzzball-muck/fuzzball/jobs/493378581#L365
https://cirrus-ci.com/task/6399888666394624

### DIFF
--- a/src/fbsignal.c
+++ b/src/fbsignal.c
@@ -224,12 +224,16 @@ wall_status(char *s)
 void
 sig_reconfigure(int i)
 {
-    wall_status("Configuration reload requested remotely.");
-    if (!reconfigure_ssl()) {
-        wall_status("Certificate reload failed!");
-    } else {
-        wall_status("Certificate reload was successful.");
-    }
+	#ifdef USE_SSL
+	wall_status("Certificate reload requested remotely.");
+	if (reconfigure_ssl()) {
+		wall_status("Certificate reload was successful.");
+	} else {
+		wall_status("Certificate reload failed!");
+	}
+	#else
+	wall_status("TLS certificate reload requested, but this server wasn't compiled to use TLS.");
+	#endif
 }
 
 /*


wyld-sw:
Thanks for catching the need for checking USE_SSL. As reloading configuration would eventually be used for more than just SSL, the messages can stay like they are for now.

I'll include USE_SSL in my next commit.

nyanster:
Could we have separate signals for reloading that future configuration and for re-loading TLS certs?  Unless certs become part of that future configuration.

wyld-sw:
I assumed that reloading configuration meant everything that could be reloaded (which is moot, since there's only one thing at the moment). Your question is a valid one, and would think SIGHUP could handle everything. What do we think from a MUCK admin perspective and server admin perspective, @tanabi and @digitalcircuit ?

tanabi:
> I assumed that reloading configuration meant everything that could be reloaded (which is moot, since there's only one thing at the moment). Your question is a valid one, and would think SIGHUP could handle everything.

I would say reload everything which would be (I think) the expected result.  And honestly it should be pretty harmless if the only thing that changed is the SSL cert.

nyanster:
> I assumed that reloading configuration meant everything that could be reloaded (which is moot, since there's only one thing at the moment). Your question is a valid one, and would think SIGHUP could handle everything. What do we think from a MUCK admin perspective and server admin perspective, @tanabi and @digitalcircuit ?

That makes sense.  I was thinking in terms of configuration files, etc.  But you're right.  I still think keeping it worded as "certificate" makes sense for now until more is added to the function, but, I'm fine either way, now.

nyanster:
I'll send an alternate diff in a moment.

wyld-sw:
Oops, already done. Let me know if there's still an issue. And we'll see if we can find additional functions for sig_reconfigure() :)

PR #319, opened 2019-02-16
Ip forward

hyena:
This is similar to the X-Forwarded-For request header in HTTP.
The purpose is to provide administrators running a FORWARDED
aware proxy or webclient accurate hostnames in WHO*.

A client must indicate their desire to use this option with
IAC WILL FORWARDED and receive a IAC DO FORWARDED in response,
after which they may at any time send a sequence like:
IAB SB FORWARDED <hostname> IAC SE.

This is only supported for gateways running on the localhost.

### DIFF
--- a/include/interface.h
+++ b/include/interface.h
@@ -66,6 +66,7 @@ typedef enum {
 #define TELNET_SE	240
 
 #define TELOPT_STARTTLS	46
+#define TELOPT_FORWARDED  113 /* non-standard; for gateways */
 
 struct text_block {
     size_t nchars;
@@ -104,6 +105,12 @@ struct descriptor_data {
     telnet_states_t telnet_state;
     int telnet_sb_opt;
     int short_reads;
+
+    /* Fields related to ip-forwarding, to support websocket gateways  */
+    int forwarding_enabled;
+    char *forwarded_buffer;
+    int forwarded_size;
+
     time_t last_time;
     time_t connected_at;
     time_t last_pinged_at;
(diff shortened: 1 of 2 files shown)


hyena:
In order to test this, you'll need a proper gateway. Morticon is preparing one I understand. Or you can test with my very barebones solution RatChat: https://github.com/hyena/muck-websocket.

Some more notes on the solution:
  1. This only supports gateway connections from the localhost. This is intentional as the gateway written by Morticon doesn't support SSL connections when last I heard.
  2. As written localhost connections (i.e. those from 'localhost' or '127.0.0.1') as well as ones forwarded through the gateway are treated as secure in WHO output and in MUF (`descrsecure?`). This is under the belief that both will be using the local loopback.
  3. The behavior from the localhost being shown as 'secure' regardless of whether it's using SSL or not is one I consider a little controversial and it's subject to discussion.

Testing was done by connecting to the server running the `minimal.db` via telnet on the localhost on a non-SSL port and then again via my RatChat gateway modified to advertise the hostname as `"this-is-a-hostname-from-the-gateway"`. The output of `WHO*` was as follows:

```
Player Name                Location     On For Idle   Host
One(#1)                    [    46]      00:00   0s @ this-is-a-hostname-from-the-gateway(36450)
One(#1)                    [    46]      02:11   1m @ 127.0.0.1(34430)
2 players are connected.  (Max was 2)
```

The top connection from `One` is the one connecting through the gateway. Note the hostname and the fact that it's shown as secure. The bottom connection from `One` is connected via non-SSL telnet from the localhost. Note that it's shown as secure as well.

PR #328, opened 2019-02-27
Move user docs to docs folder

nyanster:
The intention of this diff is to move all the documentation that humans (and fluffy animals) might want to read to a single top-level folder in the tarball, so they're easy to find.  Then, only files that fuzzball will read are copied back into game/data
For example, none of the .raw or .html files are read by fuzzball.  Thus, they don't get put into game/data
Since the .raw don't get installed to INSTALL_HELPDIR later on (implying they aren't meant to be read by documentation-seeking humans) but also aren't read by fuzzball, I treat them as source files and moved them to src

Unrelatedly, use autoconf's docdir folder instead of building the path ourselves off of datadir, for INSTALL_HELPDIR.  The folder remains the same.  This should improve things for as long as we use autoconf, in terms of user customization.

PLEASE tell me what you think of this!  I realise this is somewhat subjective, as one other option is to just say in the README "Hey, all documentation is in game/data"  But I think this is a more useful organization of the docs, especially as some stuff in game/data is config files rather than documentation.


tanabi:
This change makes good sense to me.  It isn't obvious that the HTML versions of those help files exist unless you are looking for them.  We could also theoretically fold the doxygen generated files under docs as well (I currently have them being generated to api-docs) though we may like to keep those separate since the audience for those docs is separate.

nyanster:
Oh.  I'm gonna have to change something here.  The Windows AppVeyor artifacts include no documentation at all now.
I have no access to Windows though and have no idea how to program on it though. :|  but I'll try

nyanster:
I think this is good to commit.  Windows artifacts lose documentation, but Windows shouldn't hold Fuzzball back.  If anyone cares about Windows support, they can fix it.

nyanster:
By the way, check out these:
https://docs.microsoft.com/en-us/windows/wsl/about
https://en.wikipedia.org/wiki/Windows_Subsystem_for_Linux
Of course, Cygwin and MinGW still exist too.

wyld-sw:
I have installed fuzzball on WSL and it seems to work well. Didn't try connecting from the outside though.

nyanster:
> I have installed fuzzball on WSL and it seems to work well. Didn't try connecting from the outside though.

Searching online shows it's possible with the right firewall settings.  I haven't confirmed this, though.  Let us know! :D

Also, the failing Travis check is just because it was cancelled.  It succeeded earlier.

wyld-sw:
Does makefile.win need to be updated to copy the files for Windows?

tanabi:
I have windows 7 in a VM with the M$ dev tools set up. Not necessarily offering to be the windows "test bitch" because I have very little interest in the work.  But, if you introduce an incompatibility with Windows, feel free to make a "come back and fix it later" ticket and I'm sure @wyld-sw or I will poke it with a stick later :)

nyanster:
> Does makefile.win need to be updated to copy the files for Windows?

Either that or appveyor.yml, yes, if a native Windows build (rather than WSL) is to be supported by Fuzzball itself rather than by a downstream derivative or fork. (Like a FreeBSD -> HardenBSD or Debian -> Ubuntu sort of relationship)

> I have windows 7 in a VM with the M$ dev tools set up. Not necessarily offering to be the windows "test bitch" because I have very little interest in the work. But, if you introduce an incompatibility with Windows, feel free to make a "come back and fix it later" ticket and I'm sure @wyld-sw or I will poke it with a stick later :)

Okidoki!

nyanster:
Ping.  I'd really like to get this documentation reorganization in.  What's the current outlook?

wyld-sw:
As long as we put the command to copy the files for Windows in makefile.win, I'm fine with merging this.

nyanster:
> As long as we put the command to copy the files for Windows in makefile.win, I'm fine with merging this.

Okidoki!  And thank you so much for merging this!!
I made an issue to track that:
https://github.com/fuzzball-muck/fuzzball/issues/330

PR #332, opened 2019-03-04
Enable yield/overt flags by default

nyanster:
For now, just by flipping the tune.  This gives users, particularly builders, far more command over the rooms they build.
If this works out alright and no-one dislikes it...
Future directions: maybe remove the tune and enable Y and O flags for all MUCKs.
TODO: get all default flags to actually show up with "[default]" in at-tune.  I think this merely means placing a percent symbol in front of each tune in each db.


wyld-sw:
Accepted. I don't know of any reason why we couldn't remove the sysparm in a future commit.

nyanster:
@wyld-sw 
Thanks :D

PR #333, opened 2019-03-04
Update last few version numbers

nyanster:
I wonder if there's a simpler way to do this.  Well, I guess for now, version numbers can be dealt with lazily, as this is -master.

### DIFF
--- a/docs/help.txt
+++ b/docs/help.txt
@@ -1,4 +1,4 @@
-           MUCK Reference Manual for Muck2.2fb7.00b1.20190227a
+           MUCK Reference Manual for Muck2.2fb7.00b1.20190304a
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/docs/man.txt
+++ b/docs/man.txt
@@ -1,4 +1,4 @@
-           MUF Reference Manual for Muck2.2fb7.00b1.20190227a
+           MUF Reference Manual for Muck2.2fb7.00b1.20190304a
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/docs/mpihelp.txt
+++ b/docs/mpihelp.txt
@@ -1,4 +1,4 @@
-           MPI Reference Manual for Muck2.2fb7.00b1.20190227a
+           MPI Reference Manual for Muck2.2fb7.00b1.20190304a
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/include/config.h
+++ b/include/config.h
@@ -15,7 +15,7 @@
 
 #include "autoconf.h"
 
-#define VERSION "Muck2.2fb7.00b1.20190227a"
+#define VERSION "Muck2.2fb7.00b1.20190304a"
 
 /************************************************************************
    Administrative Options 
--- a/win32/autoconf.h
+++ b/win32/autoconf.h
@@ -83,7 +83,7 @@
 #define PACKAGE_NAME "fbmuck"
 
 /* Define to the full name and version of this package. */
-#define PACKAGE_STRING "fbmuck 7.00b1.20190227a"
+#define PACKAGE_STRING "fbmuck 7.00b1.20190304a"
 
 /* Define to the one symbol short name of this package. */
 #define PACKAGE_TARNAME "fbmuck"
@@ -92,7 +92,7 @@
 #define PACKAGE_URL ""
 
 /* Define to the version of this package. */
-#define PACKAGE_VERSION "7.00b1.20190227a"
+#define PACKAGE_VERSION "7.00b1.20190304a"
 
 /* The size of a `int', as computed by sizeof. */
 #define SIZEOF_INT 4


tanabi:
We could do something similar to mkversion to set version in a common place (either in a global variable in version.c or we could have mkversion create a .h file with the version numbers).  For the MUF and MPI man pages, I'd say just take the version number out of there ... that just doesn't strike me as a maintainable or even good place to put version info (just IMO).

nyanster:
> We could do something similar to mkversion to set version in a common place (either in a global variable in version.c or we could have mkversion create a .h file with the version numbers). For the MUF and MPI man pages, I'd say just take the version number out of there ... that just doesn't strike me as a maintainable or even good place to put version info (just IMO).

Well, that's what autoconf already does, yeah.  Things just aren't as convenient as they could be. xP  They either have to go into commits, or go into separate commits... the version number changes, that is.
Perhaps true about the man pages.  (though I think them having release version numbers is OK, they don't need -master version numbers, and would be fine with no version number at all, too)
Since autoconf already does this, the problem could just be me not installing and learning autoconf and git. :P

tanabi:
> Well, that's what autoconf already does, yeah. Things just aren't as convenient as they could be. xP They either have to go into commits, or go into separate commits... the version number changes, that is.
> Perhaps true about the man pages. (though I think them having release version numbers is OK, they don't need -master version numbers, and would be fine with no version number at all, too)
> Since autoconf already does this, the problem could just be me not installing and learning autoconf and git. :P

I think that mkversion is built and run regardless of autoconf which is why I pointed it out as a possibility.  Though you're right that autoconf is "the right way to do it".  I don't really know autoconf that well, either, and it seems a little like a scary monster. :)

nyanster:
> I think that mkversion is built and run regardless of autoconf which is why I pointed it out as a possibility. Though you're right that autoconf is "the right way to do it". I don't really know autoconf that well, either, and it seems a little like a scary monster. :)

Oh!  Good point!  While we have mkversion, and while we ARE seeking to remove autoconf, moving the version number from configure.ac to somewhere in mkversion is a **very good idea**.

wyld-sw:
VERSION is #defined in config.h - I'm completely file with the detailed version number from the configure process.

Should there be a MAJOR_VERSION and MINOR_VERSION so that things like help/man/mpi docs don't show the whole thing - or should we just decide what the versioning policy should be - and where to show what?

I just saw version.c and it appears the compilation time is still in there. It's possible version.c was updated before (this file is not in the repository and is generated - by mkversion - from version.tpl)

nyanster:
> Should there be a MAJOR_VERSION and MINOR_VERSION so that things like help/man/mpi docs don't show the whole thing - or should we just decide what the versioning policy should be - and where to show what?

We should decide something on versioning policy.  Though, your idea is interesting.  Creating a  COMMITDATE or commit-related variable separate from the version variable sounds good.
I myself don't think versioning -master is important.  If someone is running -master rather than a release, they should be expected to actually be on the newest commit from -master when reporting/duplicating a bug.  THOUGH, this leads to a scenario where competent MUCK wizzes would be required when checking bugs on a MUCK.  Though...  I assume any MUCK wiz running -master is willing to do updates.
I did put in the date because I was asked, but, is it truly necessary?  Genuine, non-rhetorical question there.
> I just saw version.c and it appears the compilation time is still in there. It's possible version.c was updated before (this file is not in the repository and is generated - by mkversion - from version.tpl)

make clean? :P  I'm pretty sure I removed the code to even put a compilation time in there.

wyld-sw:
You're right, I was looking at an older copy!

wyld-sw:
No, I don't feel the date is necessary (I had given it as an example of something a number that - more often than not - increases).

Feel free to come up with another scheme.

nyanster:
> No, I don't feel the date is necessary (I had given it as an example of something a number that - more often than not - increases).
> 
> Feel free to come up with another scheme.

Well, I meant removing that number entirely, rather than any thought on the format of it. xP
I realize it is the only way to know how old a -master-running MUCK is, but at the same time think it's the MUCK admin's responsibility to keep -master at -master if they choose not to run a release instead.
I won't fuss either way. |3

wyld-sw:
Yeah, I agree we don't need it there. Another scheme, as in versioning policy. Sorry for the confusion.

nyanster:
How about switching from 7.00b1 to 7.00b1+master, and maybe switching to 7.00b2-master close to b2 release, then switching to 7.00b2 for a **single commit only** which gets tagged as release, then moving to 7.00b2+master?  The word master might be optional if you wanna just use + and - like Korn Shell version numbers do.  (though, the + and - are used in **releases** for ksh93, I think the concept is easily adapted to here)

nyanster:
> How about switching from 7.00b1 to 7.00b1+master, and maybe switching to 7.00b2-master close to b2 release, then switching to 7.00b2 for a **single commit only** which gets tagged as release, then moving to 7.00b2+master? The word master might be optional if you wanna just use + and - like Korn Shell version numbers do. (though, the + and - are used in **releases** for ksh93, I think the concept is easily adapted to here)

Actually hmm, yeah, I'd definitely use the "master" part to make it more obvious that it's not a release.

wyld-sw:
I agree with that; it also allows the larger feature request version codes to specify the branch they're on.

PR #335, opened 2019-03-04
Include standard C headers

nyanster:
I think that's all, for getting rid of standard C header ifdefs.
Potential TODO: get rid of posix header ifdefs?  unistd.h and dirent.h are what I'm thinking of.  From what I understand, fuzzball wouldn't work without them...  But I'm keeping that for a separate diff, since those'll have to go into a ifndef WIN32 zone.
Also should check each define in autoconf.h.in for whether we're still using them or not.  Some of them are definitely unused and might never be used again.

This diff includes stdint.h, gets rid of some mediocre typedefs related to that, and gets rid of a few things no longer used in autoconf.h.in

### DIFF
--- a/include/autoconf.h.in
+++ b/include/autoconf.h.in
@@ -52,12 +52,6 @@
 /* Define to 1 if you have the `random' function. */
 #undef HAVE_RANDOM
 
-/* Define to 1 if you have the <strings.h> header file. */
-#undef HAVE_STRINGS_H
-
-/* Define to 1 if you have the <string.h> header file. */
-#undef HAVE_STRING_H
-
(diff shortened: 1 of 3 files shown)


tanabi:
@nyanster I haven't verified this, but the POSIX ifdef may be to support Windows.  You are right that the MUCK won't work without those headers, but I think Windows puts those system calls in a different header file if i remember right  (For Windows, I think the posix stuff is almost entirely all in the winsock.h file or something like that ....).  Anyway, if we were to remove these ifdefs we should check the windows API first to make sure we're not breaking something first.

tanabi:
Sounds like you already said that in your comment and I'm sleepy brained, don't mind me, doo da doo :)

PR #339, opened 2019-03-04
Change to new versioning scheme

nyanster:
This is something that'll have to be changed FAR less often than one that includes the date, while still telling users between which releases a -master-running MUCK is on.
I'm happy with this.


nyanster:
Potential TODO:
Put that configure.ac comment about versioning into config.h too or instead.

nyanster:
Actually, lemme do that now before it's merged

PR #341, opened 2019-03-04
Use normal commands as examples in help

nyanster:
Have examples use say and pose rather than " and : shorthands.  Shorthands are still documented.
Also correct typos.

### DIFF
--- a/docs/help.txt
+++ b/docs/help.txt
@@ -1,4 +1,4 @@
-           MUCK Reference Manual for Muck2.2fb7.00b1+master
+            MUCK Reference Manual for Muck2.2fb7.00b1+master
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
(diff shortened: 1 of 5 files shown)


tanabi:
I disagree with his change, because I think more people use the short-hand than the long-form.

I would suggest showing examples of *both* instead so its clear to new users (because these help files are targeted to the newest of the new!) how it works.

nyanster:
@tanabi 
I get what you mean, but I think the short forms confuse users.  I've known multiple people who always do "foobar" instead of "foobar despite the fact that their error is visible in their own say.
I'll edit this to include both.  One moment~

nyanster:
It's ready for merge now.  It now includes shorthand examples too.

tanabi:
Perfect!  Thanks @nyanster

nyanster:
In general, I'm not a fan of the fact that the short forms allow removing the space between the command and the message.  Even in interactive fiction, "examine door" is shortened to "x door" rather than "xdoor"
uwu
but yeah, no prob tanabi :D

tanabi:
> In general, I'm not a fan of the fact that the short forms allow removing the space between the command and the message. Even in interactive fiction, "examine door" is shortened to "x door" rather than "xdoor"
> uwu
> but yeah, no prob tanabi :D

Personally, I never use the short form for say -- I always type "say".  I've been doing that for almost 30 years.  People who watch over my shoulder always mock me for it :)

Pose, however, I use the short form... I think the lack of space is mostly because in a pose, sometimes you don't have a space.

```
:'s biscuits are tasty.
```

for instance.  Putting a space makes the pose feel clunky IMO.  Though I never could get used to the short form say, and I think requiring a space in that case is perfectly reasonable, but the genie is way way way out of the bottle on this one so its never going to change despite the oddity of it :D

nyanster:
> the genie is way way way out of the bottle on this one so its never going to change despite the oddity of it :D

truedat. :B

aidanPB:
I don't think it would be a disaster to require spaces for everything except `"` and `:`. I certainly never leave them out for anything else. (Mind, I hardly ever use any builtin shorthand forms other than `"` and `:`.)

On the other hand, I also don't think it would be worth the effort to code in the exception.

nyanster:
> On the other hand, I also don't think it would be worth the effort to code in the exception.

Well, the thing is that the exception already exists.  In include/game.h we see SAY_TOKEN and POSE_TOKEN are unlike any other command, among some other special commands.
That said, I never knew about the ```delimiter``` command until I just looked it up now.  Interesting.

wyld-sw:
Slightly related: I think you can create actions that don’t require the space by setting the exit (X)PRESS. I’ve never done it, so I may have dreamt some of that :)

PR #345, opened 2019-03-06
Regenerate minimaldb

nyanster:
Did this by wiping sysparms from db manually and saving what fuzzball itself saves back to db.
Does anyone know what the now-gone sysparm is?  I remember some have been recently removed.  (and the number is 157 instead of 158 now...)

I manually deleted all but one sysparm in minimaldb, changed 158 to 1, and fuzzball was able to start.  I then started, did at-shutdown, and voila, minimaldb was made again.
I want to figure out how to make the smallest minimaldb possible, but UNTIL THEN, this at least makes all sysparms show up as [default] when they are default.  That is a good success.

I do realise some extra props are at the bottom.

Onward to trueminimaldb!

### DIFF
--- a/dbs/minimal/data/minimal.db
+++ b/dbs/minimal/data/minimal.db
@@ -188,12 +188,6 @@ Room Zero
 *Props*
 _/de:2:You are in Room Zero. It's very dark here.
 _/osc:2:is briefly visible through the mist.
-_sys/dumpinterval:3:14400
-_sys/lastdumptime:3:1551890386
-_sys/max_connects:3:1
-_sys/maxpennies:3:10000
-_sys/shutdowntime:3:1551890386
-_sys/startuptime:3:1551890371
 *End*
 -1
 -1
--- a/dbs/minimal/data/minimal.db
+++ b/dbs/minimal/data/minimal.db
@@ -1,165 +1,164 @@
 ***Foxen9 TinyMUCK DUMP Format***
 2
 1
-158
-autolook_cmd=look
-cpennies=Pennies
-cpenny=Penny
-pennies=pennies
-penny=penny
-description_default=You see nothing special.
-dumpdone_mesg=## Save complete. ##
-dumping_mesg=## Pausing to save database. This may take a while. ##
-dumpwarn_mesg=## Game will pause to save the database in a few minutes. ##
-file_connection_help=data/connect-help.txt
-file_credits=data/credits.txt
-file_editor_help=data/edit-help.txt
-file_help=data/help.txt
-file_help_dir=data/help
-file_info_dir=data/info/
-file_log_cmd_times=logs/cmd-times
-file_log_commands=logs/commands
-file_log_gripes=logs/gripes
-file_log_malloc=logs/malloc
-file_log_muf_errors=logs/muf-errors
-file_log_programs=logs/programs
-file_log_sanfix=logs/sanfixed
-file_log_sanity=logs/sanity
-file_log_status=logs/status
-file_log_stderr=logs/fbmuck.err
-file_log_stdout=logs/fbmuck
-file_log_user=logs/user
-file_man=data/man.txt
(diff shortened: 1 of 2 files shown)


nyanster:
For context, trueminimaldb would be the smallest db fuzzball actually starts with and doesn't complain about.

nyanster:
muf_string_math is the removed sysparm! :D

wyld-sw:
It looks like the new "description_default" sysparm may be missing in the regenerated file. I may have done something weird that suppresses it. I can check, if no-one beats me to it.

And previous versions of the database format did not include sysparms in the file. It basically did a `@tune load` during startup if I remember right.

Maybe a new database format could restore support for an auto-loaded sysparm file - among several other things.

nyanster:
> And previous versions of the database format did not include sysparms in the file. It basically did a `@tune load` during startup if I remember right.
> 
> Maybe a new database format could restore support for an auto-loaded sysparm file - among several other things.

I think they're fine being in the db file.

wyld-sw:
I do agree with removing minimal.db from game/. It exists mainly to support more experienced users who have a "routine" to set up a test MUCK. I don't think there's much need for it at this point.

nyanster:
what IS weird is they're allowed to be saved and loaded separately
Is that an artifact of them being in a different file previously, or?
And okidoki, I can write the documentation for removing it.  Lemme do that and the removal in the same commit.

wyld-sw:
Re: sysparms in the db file: Ok, I misunderstood your intention with trueminimal. Good to know we're on the same page! :)

Re: saved and loaded sysparms. I would consider it an artifact, but I can also come up with a (possibly shaky) use-case: loading and saving settings file could help when trying to synchronize sysparms between MUCKs?

nyanster:
> Re: sysparms in the db file: Ok, I misunderstood your intention with trueminimal. Good to know we're on the same page! :)
> 
> Re: saved and loaded sysparms. I would consider it an artifact, but I can also come up with a (possibly shaky) use-case: loading and saving settings file could help when trying to synchronize sysparms between MUCKs?

Oof, that is shaky.  I think a less shaky (but still very shaky) use case is the idea of saving sysparms without having to go through saving the whole universe...
How would you feel about removing the at-tune save and load commands?  I don't think I can do that, though, so I won't push either way. :P

nyanster:
Regular db saves DO save them, don't they?  Not just at-shutdown, right?

wyld-sw:
yeah, db saves call db_write -> db_write_header -> tune_save_parms_to_file

I'm leaning toward saying yes to removing `@tune load` and `@tune save`. @tanabi, @aidanPB, any thoughts on that?

tanabi:
@wyld-sw @nyanster nyanster you are right those functions / the ability to save tune to a different file is an artifact of past MUCK versions.  Definitely pre FB6, maybe even pre FB5 ... I do remember back when tune parameters were in a separate file (I even remember when you had to edit that file manually!) but they gradually made their way into the FB database.

I believe the only reason for @ tune load would be to support an ANCIENT DB going to modern FB.  I'm not entirely sure how backwards compatible FB is or how much this matters, but that is your real use case for tune load.  If, for instance, you ran a MUCK back in the early 90's and wanted to bring it back (I actually have several DB's from that era still -- a couple years a go I even fired one back up again to extract some building from it -- so it can happen!)

@ tune save has no purpose at all any more.

So what's my answer?  Definitely remove @ tune save.  Consider removing @ tune load; I think, as a manual command, it makes little sense and we can remove the command, but the functionality of being able to load tune parameters from a separate file should probably remain "under the hood".  This is trivial to do due to how tune-load is structured; the code overlap with load-from-file vs. load-from-db is near 100% so we can certainly remove the @ tune save/load switches while not breaking separate-file load.

wyld-sw:
Removed `@tune load` partially and `@tune save` completely, via 178e6b5.

PR #352, opened 2019-03-07
Regenerate basedb and advancedb

nyanster:
Basically same as here:
https://github.com/fuzzball-muck/fuzzball/pull/345
https://github.com/fuzzball-muck/fuzzball/commit/f04a4b45e67e699903ad8004793fdc3754158d3f
There might me more to do in the future.  But it brings the "default value" feature into the db, for sysparms.
Something in advancedb as for stringslib and lib-stackrng changes.  I wonder what it is.  I should learn the file format sometime.


wyld-sw:
Some props are generated during startup (including MUF _defs and _docs), but that's tricky to keep out of these files.

wyld-sw:
Here's some quick notes on the db format. It may include errors, and please let me know if it does. I was readying it for inclusion before db_write() in db.c, but here seems like a good place to put it for now, for discussion.

```
/* Foxen9 DB format (brief notes)
 * - feel free to relocate, reword, or change format of these docs.
 *
 * string: "***Foxen9 TinyMUCK DUMP Format***"
 * number: # of objects in db
 * number: unused value included for backward compatibility
 * number: # of sysparms
 * repeat: for each sysparm
 * string: - "(sysparm)=(val)"
 * repeat: for each object (o), highest id first
 * dbref : - id of (o) prefixed with pound sign: "#xxxx"
 * string: - object name
 * number: - id of (o)'s location
 * number: - id of first object in (o)'s contents list, or -1 (NOTHING)
 * number: - id of first object in (o)'s "next" list, or -1 (NOTHING)
 * number: - bit field representation of all non-internal flags for (o)
 * number: - timestamp of (o)'s creation
 * number: - timestamp of (o)'s last use
 * number: - count of uses of (o)
 * number: - timestamp of (o)'s last modification
 * string: - "*Props*"
 * repeat: - for each property of (o)
 * string: -- (path):(type):(val)
 * string: - "*End*"
 * cond  : - if (o)'s type is THING
 * number: -- id of (o)'s home
 * number: -- id of first object in (o)'s exits list, or -1 (NOTHING)
 * number: -- id of (o)'s owner
 * cond  : - else if (o)'s type is ROOM
 * number: -- id of (o)'s dropto
 * number: -- id of first object in (o)'s exits list, or -1 (NOTHING)
 * number: -- id of (o)'s owner
 * cond  : - else if (o)'s type is EXIT
 * number: -- count of (o)'s destinations
 * repeat: -- for each exit (e) located on (o)
 * number: --- id of (e)
 * number: -- id of (o)'s owner
 * cond  : - else if (o)'s type is PLAYER
 * number: -- id of (o)'s home
 * number: -- id of first object in (o)'s exits list, or -1 (NOTHING)
 * string: -- (o)'s password (md5)
 * cond  : - else if (o)'s type is PROGRAM
 * number: -- id of (o)'s owner
 */
```

There's a bit of redundancy. The specific lines for rooms and things are conceptually the same, since both drop-tos and homes are associated the same way - though they're different variables in the object structures. We have plenty of time to address this in both a new db format and the code structures.

Issue #354, opened 2019-03-07
Possibly using stdbool.h


nyanster:
How would you guys feel about using the standard C header stdbool.h?  While reading through the code, I often think, "mmm, it'd be nice if a more readable way was used to save memory"
Reading up on stdbool, I see it defines bool as _Bool, and in C99 and in the latest version of C, a _Bool only needs to be big enough to fit 0 and 1. (that's in section 6.2.5, list item number 2, for C99)  That could preserve memory savings while being far more readable than some current practices.
I was reminded by how we store all flags with those long hexadecimal numbers... though I realise those might not be what get replaced, the idea extends to other things too.

References:
C99:
http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
C17 (C18?):
https://web.archive.org/web/20181230041359if_/http://www.open-std.org/jtc1/sc22/wg14/www/abq/c17_updated_proposed_fdis.pdf
POSIX:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/stdbool.h.html

possible downside: the code uses "bool" within other variable/function names to mean boolean expressions (boolexps) and other fuzzball-related-booleans.  Is there possibility for confusion?
Also, in WIN32 ifdefs in .c files in src/, there are a total of four BOOL (uppercase) variables.
win32/restart.cpp also uses bool (lowercase)
bool itself though IS available, as are true and false

We can always double-check by including stdbool.h in config.h to do a CI check.

wyld-sw:
I would be fine with it, as it can increase understanding of the code.

tanabi:
> Reading up on stdbool, I see it defines bool as _Bool, and in C99 and in the latest version of C, a _Bool only needs to be big enough to fit 0 and 1. (that's in section 6.2.5, list item number 2, for C99) 

From a practical standpoint, a C bool is the same data size as a 'char' because there is no way to address something lower than a byte.  However, I think in some places we're using 'short's for bool (2 bytes) so in those places we'd be saving.  Here's a test for instance:

```
#include <stdbool.h>
#include <stdio.h>

int main(int argc, char** argv, char** envp)
{
        printf("%ld\n", sizeof(bool));
        return 0;
}
```

I like bool semantically but I don't think it really does anything for performance or memory utilization on the whole.  Anyway, I guess this is a roundabout way of saying this is a good idea though probably doesn't really have a performance or utilization.

[...]

PR #355, opened 2019-03-07
Add stdbool.h (don't merge, CI check)

nyanster:
I'm most interested in seeing how Windows builds react, as a CI check..  This IS part of the standard C library, so Windows should be okay, but...  I wonder~
I want to see how people react here
https://github.com/fuzzball-muck/fuzzball/issues/354
first though, so don't merge.  I'd also want to find a good use for it and actually use it before adding it, of course.

### DIFF
--- a/include/config.h
+++ b/include/config.h
@@ -157,6 +157,7 @@
 #include <limits.h>
 #include <signal.h>
 #include <stdarg.h>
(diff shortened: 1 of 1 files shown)


nyanster:
Woah, works everywhere, _including on Windows_.  Nice.

PR #362, opened 2019-03-14
Remove do_unlink_quiet

nyanster:
I think this is OK since there's no at-unlinkquietly command for users.
TODO done. :3


nyanster:
Alternatively, to avoid the sort of confusion I had, I wonder if _do_unlink could be renamed to not look like a do_XYZ function.  Maybe unlink_thing or _unlink_thing or something like that.
Not necessary though.

PR #363, opened 2019-03-15
Finished compile.c and various doc corrections

tanabi:
I still don't know how some of this works.

Please note that the GIT file diff browser seems to get kind of confused in places on this PR.  It seems like it works out correctly though, just some places it looks a little strange.  You'll see what I mean I'm sure. (Actually I think the reason for this is I didn't merge changes from master before submitting the PR -- I've merged in master and I think it looks more correct)


nyanster:
A new warning was introduced by this PR:
https://travis-ci.org/fuzzball-muck/fuzzball/jobs/506618678#L319
https://github.com/fuzzball-muck/fuzzball/pull/363/files#diff-3dc78d5c2e3afa7dcd2963e0adc5e9ddR4763
https://clang.llvm.org/docs/DiagnosticsReference.html#wempty-body
I find this odd because the code looks unchanged other than for indentation. (though, compilers are known to use indentation to determine warnings)

nyanster:
> A new warning was introduced by this PR:
> https://travis-ci.org/fuzzball-muck/fuzzball/jobs/506618678#L319
> https://github.com/fuzzball-muck/fuzzball/pull/363/files#diff-3dc78d5c2e3afa7dcd2963e0adc5e9ddR4763
> https://clang.llvm.org/docs/DiagnosticsReference.html#wempty-body
> I find this odd because the code looks unchanged other than for indentation. (though, compilers are known to use indentation to determine warnings)

Ah, it's possibly because your indentation does indeed imply the lower line is part of the for loop when it's not.

src/set.c
```diff
@@ -1132,7 +1132,7 @@ do_propset(int descr, dbref player, const char *name, const char *prop)
  * is no object name, then the registration target is cleared.  Otherwise,
  * object is registered to the given name.
  *
- * See help @register on the MUCK for a more consise explanation :)
+ * See help @register on the MUCK for a more consice explanation :)
```
nyanster: consice -> concise
;w;

src/move.c
```diff
@@ -1140,7 +1140,14 @@ do_recycle(int descr, dbref player, const char *name)
...
                     }
 
-                    /* FIXME: This is a workaround for bug #201633 */
+                    /*
+                     * @TODO This is a workaround for bug #201633
+                     *
+                     * This bug predates sourceforge and is from a system
+                     * long gone.  MUCKs aren't dying left and right due
+                     * to this, so I'm inclined to remove this code block
+                     * now.
```
nyanster: The workaround might be what causes MUCKs not to die, though? :P  No idea.

nyanster:
This time, I guess I'll offer fixes as a pull request instead of as comments on the commit, and just have the PR be reviewed for OKs before I OK it for merge.

src/move.c
tanabi: No, it's not -- the if statement has an assert(0); in it.  MUCKs would be dying left and right if it WAS hitting this code block :)  That's what my comment intends to say here.

nyanster: > No, it's not -- the if statement has an assert(0); in it. MUCKs would be dying left and right if it WAS hitting this code block :) That's what my comment intends to say here.

Oh!  I see.  Thanks for the enlightenment. :D

nyanster: Though, isn't it also possible NDEBUG can sometimes be defined (see autoconf.h.in, config.h) causing assert() to do nothing?

wyld-sw: NDEBUG causes assert() to be disabled - which is the usual state of the MUCK (--enable-debug I think causes this not to be #defined, I think).

I think I have that logic right.

nyanster: > NDEBUG causes assert() to be disabled - which is the usual state of the MUCK (--enable-debug I think causes this not to be #defined, I think).
> 
> I think I have that logic right.

Correct. I just compiled and confirmed this.

wyld-sw: Yes, in config.h - Thanks for the confirmation.

nyanster: This was marked as resolved for some reason (which causes this to be nearly hidden on the issue's webpage :c ) even though what we've determined contradicts the conclusion of the comment in question.

nyanster: This was marked as resolved (which causes this to be nearly hidden on the issue's webpage :c ) even though what we've determined contradicts the conclusion of the comment in question.

wyld-sw: @nyanster, it was a mistake. I unresolved it immediately.

tanabi: @nyanster @wyld-sw So what do we want to do then?  Just leave it as a mystery?  Put a log message in there and see if it shows up anywhere?  The latter would require cooperative MUCKs -- I mean Hope Island would cooperate but we probably need a larger sample size.

wyld-sw: I would prefer to leave it in (with a log message as you suggest), at least temporarily. I've never seen this message, but until we determine that dequeue_prog always works as expected, I don't think we can remove it.

tanabi: Actually on further thought ... this is SUCH a rare case I'm not sure if we'll see it in the wild.  We'd probably have to test it on a dev MUCK.  How often does a live MUCK @ recycle programs?  I would say hardly ever.

nyanster: > Actually on further thought ... this is SUCH a rare case I'm not sure if we'll see it in the wild. We'd probably have to test it on a dev MUCK. How often does a live MUCK @ recycle programs? I would say hardly ever.

o, true

src/set.c
tanabi: Someday I'll spell that right.  It'll be in my next PR.  Or I guess your PR if you fix it first!

tanabi:
> > A new warning was introduced by this PR:
> > https://travis-ci.org/fuzzball-muck/fuzzball/jobs/506618678#L319
> > https://github.com/fuzzball-muck/fuzzball/pull/363/files#diff-3dc78d5c2e3afa7dcd2963e0adc5e9ddR4763
> > https://clang.llvm.org/docs/DiagnosticsReference.html#wempty-body
> > I find this odd because the code looks unchanged other than for indentation. (though, compilers are known to use indentation to determine warnings)
> 
> Ah, it's possibly because your indentation does indeed imply the lower line is part of the for loop when it's not.

Good catch ... fell into the trap that this compile warning is detecting.  I will remove the incorrect indent and see if that fixes the issue.  (Unless you beat me to it).  I must be clicking the wrong things in travis because I checked it last night and didn't see this. :P

src/move.c
tanabi: I'm inclined to make this a separate issue, put the issue number in the TODO comment, and try to sort this out on a dev MUCK.  Even then, we may not be able to reproduce it, but we can at least make a concentrated effort to do so and list our test cases out so that if this bug ever does crop up, we've got public history of it rather than a mysterious ticket number in an unknown ticket system.  If that sounds good I'll go ahead and make the ticket for it.

nyanster: > I'm inclined to make this a separate issue, put the issue number in the TODO comment, and try to sort this out on a dev MUCK. Even then, we may not be able to reproduce it, but we can at least make a concentrated effort to do so and list our test cases out so that if this bug ever does crop up, we've got public history of it rather than a mysterious ticket number in an unknown ticket system. If that sounds good I'll go ahead and make the ticket for it.

Sounds good to me.

wyld-sw: It does sound good, thanks!

PR #367, opened 2019-03-17
Fix up includes a bit, part 1 of 2 or more

nyanster:
The goal of these diffs is to get more includes into the files they belong in.  This is in preparation for using iwyu, https://include-what-you-use.org/ to help us move standard C header includes out of config.h and into the files they belong in.

These results come from me applying the automatic fixes from iwyu and seeing what broke, and fixing what broke.  (no!  this pull request does not include the automatic iwyu fixes.  those will come in a second PR)

### DIFF
--- a/include/compile.h
+++ b/include/compile.h
@@ -8,6 +8,8 @@
 #ifndef _COMPILE_H
 #define _COMPILE_H
 
+#include "config.h"
+
 /*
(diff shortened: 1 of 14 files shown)


nyanster:
But yeah.  all these config.h includes are merely for the dbref typedef
I'll go and split config.h into chunks sometime soon though, so they'll all become fbmuck.h includes.

PR #369, opened 2019-03-17
CI check (don't merge)

nyanster:
I just realised that iwyu makes certain assumptions.  I need to figure out if fb will still build on Windows after...  Because I don't want to spend another hour handpicking iwyu's suggestions if the Windows builds stop building.  (by hand, I'm trying to only add standard C headers, but I've realised the local headers I remove might be relevant to the windows port)
It's entirely possible we'll only be able to make iwyu-based changes after removing autoconf and dissecting config.h, in order to have a cleaner thing to work on.

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -6,7 +6,6 @@
  * This file is part of Fuzzball MUCK.  Please see LICENSE.md for details.
  */
 
-#include <assert.h>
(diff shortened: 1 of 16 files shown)


nyanster:
oh my god i'm so relieved the appveyor build passed

PR #372, opened 2019-03-17
Add more includes to .h files

nyanster:
Removals, again, are saved for later.  Most fun should be removing includes from config.h :D  but a lot of .c files and .h files will also have stuff to remove
But removals need to be done far more carefully.  Perhaps not even in a single day.


nyanster:
> Looks straight forward enough!

It was a lot better than the pull request before it for .c files...
xP

PR #386, opened 2019-04-02
use -lm to link math library

nyanster:


nyanster:
Awesome.  This should be good to merge.

PR #390, opened 2019-04-02
Add documentation to ws and status in advancedb

nyanster:
Also fix oddly broken English in a couple files.


nyanster:
Don't merge this in until I download and check that status is still blessed properly.

nyanster:
Okay.  It's all looks good to me.

PR #391, opened 2019-04-02
Remove colorisation of WIZ status

nyanster:
This is a no-no.  The status prop isn't protected in any way at all, so anyone can set themselves to have the WIZ status.  That's fine, but it shouldn't be given its own color in ws.
This should make impersonating a wiz by accident or on purpose less likely.

Documentation is also fixed in this PR.

### DIFF
--- a/dbs/advancedb/data/advancedb.db
+++ b/dbs/advancedb/data/advancedb.db
@@ -497,9 +497,9 @@ status;stat
 141
 2
 1073635051
-1554235923
-5
-1554236183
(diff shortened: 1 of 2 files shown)


nyanster:
Of note is that ```ws #setup``` fails, even as One, even with an M3 bit set.

wyld-sw:
Will fix with upcoming database-related commits. In short, `ws #setup` renames the trigger object, which requires that the program be set Wizard.

PR #393, opened 2019-04-03
Move dbref typedef to fbmuck.h

nyanster:
This is the first of likely many things that can be moved out of config.h, as it is not meant to be configured, neither by user nor by autoconf.  As the dbref typedef is pretty generic, I give it its own generic/miscellaneous header file.

The define is intentionally FBMUCK_H and not _FBMUCK_H
Having defines start with an underscore isn't a good idea.  I believe they're reserved for C implementations, and I can double-check on that if necessary.
TODO: change our current defines to not start with an _ underscore.
potential TODO: see if it's possible to have the MUCK not break when dbref is typedef'd to a different type.

### DIFF
--- a/include/commands.h
+++ b/include/commands.h
@@ -10,6 +10,7 @@
 #define _COMMANDS_H
 
 #include "config.h"
+#include "fbmuck.h"
 
 /*
  * This file is grouped alphabetically just to make it slightly easier to find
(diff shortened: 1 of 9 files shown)


wyld-sw:
You're right about the reserved-id-macro warnings. That should be a simple PR to do.

nyanster:
> You're right about the reserved-id-macro warnings. That should be a simple PR to do.

I'll do that in my next PR, then. :3  This one is good to go in.

PR #399, opened 2019-04-04
Style nitpick in interface_ssl.[ch]

nyanster:
The only thing I could find about triple-slash comments is Doxygen might pick them up, however, the things these triple-slash comments are above in interface_ssl.h, I'm not sure whether Doxygen can (or usefully can, as these are very simple descriptions) document.
@tanabi , what do you think?

If you want to convert some of these to Doxygen comments later by doubling the first asterisk, that's all fine. :D

Also, I see in the code and comments that we already have some code that uses OpenSSL 1.1.0+ features if available...  But not if not available.  Just thinking about OpenSSL version stuff.

### DIFF
--- a/src/interface_ssl.c
+++ b/src/interface_ssl.c
@@ -5,7 +5,7 @@
 #include "interface_ssl.h"
 #include "log.h"
 
-/*
+/**
  * Converts an SSL protocol version string to a version number
  *
  * Inspired by 'set_protocol_version' and 'protocol_from_string' in
@@ -26,7 +26,7 @@ ssl_protocol_from_string(const char *value)
 	return -1;
 }
 
-/*
+/**
  * Sets the minimum SSL protocol version given a version string
  *
  * If None, no change will be made to the SSL context.  If version string is invalid or unsupported
@@ -101,7 +101,7 @@ set_ssl_ctx_min_version(SSL_CTX * ssl_ctx, const char *min_version)
 	}
 }
 
-/*
+/**
  * Checks for the last SSL error, if any, recording it to the log
  *
  * If log_all is set to 0, this ignores any usually irrelevant error conditions to avoid spamming
--- a/src/interface_ssl.c
+++ b/src/interface_ssl.c
@@ -5,15 +5,15 @@
 #include "interface_ssl.h"
 #include "log.h"
 
-///
-/// Converts an SSL protocol version string to a version number
-///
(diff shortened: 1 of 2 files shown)


nyanster:
CI for this one is actually A-OK (as Cirrus for some reason failed on its own when a commit was added)

src/interface_ssl.c
```diff
@@ -96,23 +101,25 @@ set_ssl_ctx_min_version(SSL_CTX * ssl_ctx, const char *min_version)
...
-///
-/// If log_all is set to 0, this ignores any usually irrelevant error conditions to avoid spamming
-/// the log, only handling a small subset.
-///
-/// @param[in,out] d          Connection descriptor data
-/// @param[in]     ret_value  Return value from SSL function call
-/// @param[in]     log_level  Amount of logging to do according to ssl_logging_t
-/// @returns       SSL_ERROR_NONE if successful or unknown, otherwise value from SSL_get_error()
-///
+/*
```
tanabi: This needs to be /** for doxygen to read it.

src/interface_ssl.c
```diff
@@ -26,16 +26,16 @@ ssl_protocol_from_string(const char *value)
...
-/// Sets the minimum SSL protocol version given a version string
-///
-/// If None, no change will be made to the SSL context.  If version string is invalid or unsupported
-/// in this build (see SSL_PROTOCOLS), an error is logged to offer guidance on fixing the problem.
-///
-/// @param[in,out] ssl_ctx      SSL context
-/// @param[in]     min_version  Name of minimum required SSL protocol version, or "None"
-/// @returns       1 if successful, otherwise 0
-///
+/*
```
tanabi: Also needs /** for doxygen to read it

src/interface_ssl.c
```diff
@@ -5,15 +5,15 @@
...
-///
-/// Converts an SSL protocol version string to a version number
-///
-/// Inspired by 'set_protocol_version' and 'protocol_from_string' in
-/// https://github.com/openssl/openssl/blob/OpenSSL_1_1_0-stable/test/ssltest_old.c
-///
-/// @param[in] value Name of SSL protocol version
-/// @returns   Version number if found, SSL_UNSUPPORTED_PROTOCOL if unsupported, or -1 if not found
-///
+/*
```
tanabi: moar /**

nyanster:
@tanabi 
Did them. :D

I think this is OK to merge now, then.

PR #400, opened 2019-04-04
Remove unusual undef from prochelp

nyanster:
If the CI checks pass for this, this should be OK, as prochelp is used in make install, and we do make install in our checks.
I can't imagine why we need to guarantee malloc() and strdup() aren't implemented as macros, here.


nyanster:
Ah.  crt_malloc

nyanster:
I'll try to do something like this another day.  Closing this pull request.
uwu

PR #402, opened 2019-04-09
Standards check 1 of 3 (CI check don'tmerge)

nyanster:
In the matrix of
gnu99 vs c99
nothing vs -D_XOPEN_SOURCE=700
there are four possibilities.
This one checks it with gnu99 and -D_XOPEN_SOURCE=700


nyanster:
Great.  All checks passed: however, it does so with some very important warnings, some of which would be better as errors.  On to the next one.

PR #403, opened 2019-04-09
Standards check 2 of 3 (CI check don'tmerge)

nyanster:
In the matrix of
gnu99 vs c99
nothing vs -D_XOPEN_SOURCE=700
there are four possibilities.
This one checks it with c99 and nothing

We could also check with -D_POSIX_C_SOURCE=200809L and without -D_XOPEN_SOURCE=700 for an additional matrix variable but I didn't choose to, yet.  _XOPEN_SOURCE is a superset of _POSIX_C_SOURCE, so some day the latter, smaller one alone should be checked.

We've done a lot of good changes since the last time we did standard checks, hence me doing some now. :3

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,6 +13,6 @@ task:
   script:
     - ./configure --prefix="$PWD/fuzzpre" $FLAG_SSL $FLAG_MEMPROF $FLAG_DEBUG
     # Speed up CI builds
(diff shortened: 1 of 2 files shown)


nyanster:
Some of the warnings here are useful too~  Not all of them, though!  Some of them are just us using Unix features beyond C features, which is fine.

nyanster:
Actually, maybe none of these warnings are all that useful.  Well, the next check will verify this and be more useful than this one, for sure.

PR #409, opened 2019-04-11
Switch from getpagesize to sysconf

nyanster:
POSIX labeled getpagesize() as legacy a long time ago, and later removed it.  But while it was still there, but labeled as legacy, it defined it as identical to sysconf(_SC_PAGESIZE), other than for the result being a long rather than an int.

I need help with this one.  This can't be merged yet, but I don't know why things are differing.

Before:
```
Process ID: 2237
Max descriptors/process: 233379
Performed 0 input servicings.
Performed 0 output servicings.
Sent 24 messages over a socket.
Received 3 messages over a socket.
Received 0 signals.
Page faults NOT requiring physical I/O: 183
Page faults REQUIRING physical I/O: 0
Swapped out of main memory 0 times.
Voluntarily context switched 10 times.
Involuntarily context switched 0 times.
User time used: 0 sec.
System time used: 0 sec.
Pagesize for this machine: 4096
Maximum resident memory: 18064k
Integral resident memory: 1472k
```
After:
```
Process ID: 3211
Max descriptors/process: 233379
Performed 0 input servicings.
Performed 0 output servicings.
Sent 41 messages over a socket.
Received 6 messages over a socket.
Received 0 signals.
Page faults NOT requiring physical I/O: 183
Page faults REQUIRING physical I/O: 0
Swapped out of main memory 0 times.
Voluntarily context switched 15 times.
Involuntarily context switched 0 times.
User time used: 0 sec.
System time used: 0 sec.
Pagesize for this machine: 4096
Maximum resident memory: 0k
Integral resident memory: 0k
```
Despite the page size showing up correctly both times, the last two numbers turn to zero...

### DIFF
--- a/src/wiz.c
+++ b/src/wiz.c
@@ -1039,8 +1039,8 @@ do_pcreate(dbref player, const char *user, const char *password)
 void
 do_usage(dbref player)
 {
-    long psize;
 #ifdef HAVE_GETRUSAGE
+    long psize;
     struct rusage usage;
 #endif
 
@@ -1068,9 +1068,9 @@ do_usage(dbref player)
             (long) (usage.ru_maxrss * (psize / 1024)));
     notifyf(player, "Integral resident memory: %ldk",
             (long) (usage.ru_idrss * (psize / 1024)));
-#endif      /* HAVE_GETRUSAGE */
+#endif /* HAVE_GETRUSAGE */
 }
-#endif      /* NO_USAGE_COMMAND */
+#endif /* !NO_USAGE_COMMAND */
 
 /**
  * @TODO: These "topprofs" calls are almost all copy/paste.  I've noted in
--- a/src/wiz.c
+++ b/src/wiz.c
@@ -1061,8 +1061,8 @@ do_usage(dbref player)
     notifyf(player, "Swapped out of main memory %d times.", usage.ru_nswap);
     notifyf(player, "Voluntarily context switched %d times.", usage.ru_nvcsw);
     notifyf(player, "Involuntarily context switched %d times.", usage.ru_nivcsw);
-    notifyf(player, "User time used: %d sec.", usage.ru_utime.tv_sec);
-    notifyf(player, "System time used: %d sec.", usage.ru_stime.tv_sec);
+    notifyf(player, "User time used: %ld sec.", usage.ru_utime.tv_sec);
+    notifyf(player, "System time used: %ld sec.", usage.ru_stime.tv_sec);
     notifyf(player, "Pagesize for this machine: %ld", psize);
     notifyf(player, "Maximum resident memory: %ldk",
             (long) (usage.ru_maxrss * (psize / 1024)));
--- a/src/wiz.c
+++ b/src/wiz.c
@@ -1039,7 +1039,7 @@ do_pcreate(dbref player, const char *user, const char *password)
 void
 do_usage(dbref player)
 {
-    int psize;
+    long psize;
 #ifdef HAVE_GETRUSAGE
     struct rusage usage;
 #endif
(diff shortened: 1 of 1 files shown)


wyld-sw:
This is interesting. In my limited testing, the last number is always zero and the next-to-last is always non-zero - both before and after this patch. It may be an implementation detail of getrusage?

nyanster:
> This is interesting. In my limited testing (locally), the last number is always zero and the next-to-last is always non-zero - both before and after this patch. It may be an implementation detail of getrusage?

Interesting!  And quite possibly.  Both the pastes I did were from local testing on FreeBSD.  POSIX only defines ru_utime and ru_stime for rusage, with other data being optional but allowed (and undefined)
I'm not sure why changing psize from int to long would change the values of those calculations, though...  I even played around with switching 1024 to 1024L to no effect.
I wonder which values are correct, then, and what's going on.  I would like to get this change in eventually, but not if it breaks anything!! xP

nyanster:
Some unfortunate news as for GNU/Linux. :/
http://man7.org/linux/man-pages/man2/getrusage.2.html
glibc defines fields it doesn't really support, and just always has them set to 0
```
ru_idrss (unmaintained)
              This field is currently unused on Linux.
```
That's somewhat misleading.

tanabi:
I think our approach here should be twofold:

1. What are those numbers actually supposed to mean?

2. Do we care if we support them?

I think the answer to [2] is, sure, if its easy, but let's just drop them if they are going to be bad/inaccurate numbers.

The answer to [1] is, I think these are the 2 memory numbers you would see on 'ps'?  Maybe there's a better way to get them.  For the record, I don't think I've ever seen those numbers as being something accurate/meaningful -- I've only done MUCKs on Linux and Solaris.

nyanster:
Oh, mentioning Solaris...
```
The ru_maxrss, ru_ixrss, ru_idrss, and ru_isrss members of the rusage structure are set to 0 in this implementation.
```
https://docs.oracle.com/cd/E88353_01/html/E37843/getrusage-3c.html
https://docs.oracle.com/cd/E26505_01/html/816-5168/getrusage-3c.html

nyanster:
The weird thing is that ru_utime and ru_stime are showing up as 0 both on my own testing and on the dev MUCK, and they're the two things defined by POSIX to actually be useful in rusage. :c

nyanster:
> The weird thing is that ru_utime and ru_stime are showing up as 0 both on my own testing and on the dev MUCK, and they're the two things defined by POSIX to actually be useful in rusage. :c

I found the solution to this, at least.  Turns out Fuzzball is just efficient enough that it doesn't easily use up whole seconds of CPU time.  BUT timeval has both a seconds member AND a microseconds member!  With this latest patch, we now show **all** the info a POSIX-only getrusage() can provide us.  Which is just those four numbers in those two lines about user and system time. (2 timevals * 2 members per timeval = 4)

nyanster:
Bizarrely, the issue I had earlier with the last few things showing up zero... isn't there anymore, despite no further change to that code.
```
@usage
Process ID: 3102
Max descriptors/process: 233379
Performed 0 input servicings.
Performed 0 output servicings.
Sent 26 messages over a socket.
Received 9 messages over a socket.
Received 0 signals.
Page faults NOT requiring physical I/O: 184
Page faults REQUIRING physical I/O: 0
Swapped out of main memory 0 times.
Voluntarily context switched 18 times.
Involuntarily context switched 0 times.
User time used: 0 sec and 2867 usec.
System time used: 0 sec and 0 usec.
Pagesize for this machine: 4096
Maximum resident memory: 18672k
Integral resident memory: 1472k
```
I guess that meant it was some weird oddity with getrusage() itself.

That also means this patch is A-OK to go in!  Removing values likely to be inaccurate is something we can do in another patch.

PR #425, opened 2019-04-30
use appropriate types for ports and addresses

nyanster:
There are types meant for addresses and ports.  The IPv6 code was typically more correct than the IPv4 code, fixed both though.
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/netinet_in.h.html
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/inttypes.h.html

TODO:
Use a function from here:
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/arpa_inet.h.html
to translate addresses back and forth between IPv4 dotted decimal notation strings and an integers.  Currently, Fuzzball does that itself by shifting bits.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -10,6 +10,7 @@
 #include <ctype.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <inttypes.h>
 #include <stdarg.h>
 #include <stdint.h>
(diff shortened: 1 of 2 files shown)


nyanster:
Oh, I forgot something.

Issue #306, opened 2019-02-01
Segfault and Oddities when working with Autostart MCP MUF


tanabi:
Okay ... so this is a little hard to quantify, I'm still sorting out the bug.

We are running mcp-extern (which comes in the fuzzball MUFs).  Typically, when the MUCK is running after a clean start, @ ps looks like:

```
**       167   4m  15m      0  0.0     #41 cmd-@doing Sheridan     SLEEPING
**       234   --   3m      0  0.0    #652 mcp-extern HopeIsland   EVENT_WAITFOR
```

When you edit an autorun program and then compile it, it restarts the program.  In our case, its making weird duplicate processes:

```
**       PID Next  Run  KInst %CPU   Prog# ProgName   Player       
**       228   1m   5m      0  0.0    #652 mcp-extern Ikkyo        SLEEPING
**       167   3m  16m      0  0.0     #41 cmd-@doing Sheridan     SLEEPING
**       234   --   5m      0  0.0    #652 mcp-extern HopeIsland   EVENT_WAITFOR
3 events.
```

I don't know what PID 228 is.  Ikkyo is definitely not running mcp-extern, there's no way possible for him to trigger it -- I think his process used to be either a wf process or cmd-@doing, not sure which, but he had something that 'turned into' mcp-extern.  At one point, I had managed to get a total of 3 mcp-extern's on @ ps but I'm not sure how I did that.  When I @ kill'd one of the duplicate mcp-externs, it crashed the MUCK.  So this is a fun one :)

I will try to investigate this more on my test MUCK but there is something buggy happening here.

tanabi:
By the way, eventually Ikkyo's mcp-extern disappeared ...  So I guess it cleaned itself up, or the other process that was really running completed.

tanabi:
Oh, but mcp-extern turned into cmd-pose somehow:

```
**       PID Next  Run  KInst %CPU   Prog# ProgName   Player       
**       258   --  40s      0  0.0     #60 cmd-pose.m HopeIsland   EVENT_WAITFOR
1 events.
```

Something is getting mixed up, pointer wise, and I think the MUCK will inevitably crash.

tanabi:
When I open mcp-extern and recompile it, now I have 2 of these guys:

```
**       PID Next  Run  KInst %CPU   Prog# ProgName   Player       
**       258   --   2m      0  0.0     #60 cmd-pose.m HopeIsland   EVENT_WAITFOR
[...]

PR #441, opened 2019-06-16
Fix for the dreaded #306 segfault

tanabi:
@wyld-sw I know you're a merge-happy crazy-guy :)  so please read this first.

While I am .... 95% confident I've fixed this (I would be 100% if I understood the code a bit better), I may have introduced a (rare) memory leak.  I marked it with a big TODO ... There's a difference between dequeue_prog_real and dequeue_process (which underpins @ kill) that I don't really understand and I'd like your opinion on.

If you're just as clueless as me, I think we may need to do "something" to verify there's no leak.  Probably the easiest way to do this is to connect a bot to a test MUCK and have it enter @ edit on a event_wait MUF and just hit 'c' endlessly to see if the memory increases.

As I re-read this block, I'm thinking it may be something that won't be possible if we delete event queue before deleting processes, and this little if ocount < count business might just be a relic of bad logic.  It seems like something that should never happen to me, so maybe I'm stressing too much.  I just wish I understood a little better how this works, but I haven't gotten to mufevents or timequeue in the doc project yet :)

Anyway, it is probably okay to merge regardless, because a possible (extremely rare) memory leak is WAY better than a definite (extremely rare) segfault :)  And I'm about to do more MCP on Hope so having this fix will make my life better.

### DIFF
--- a/src/timequeue.c
+++ b/src/timequeue.c
@@ -925,143 +925,240 @@ get_pidinfo(int pid, int pin)
     return nw;
 }
 
-/*
- * killmode values:
+/**
+ * The implementation for the dequeue_prog define, killing processes
+ *
+ * dequeue_prog automatically injects the 'file' and 'line' parameters
+ * based on compile constants.
+ *
+ * This kills a process running on the MUCK.  Kill mode can be one of the
+ * following:
+ *
  *     0: kill all matching processes, MUF or MPI
  *     1: kill all matching MUF processes
  *     2: kill all matching foreground MUF processes
+ *
+ * file and line are used for debug purposes and should be the file/line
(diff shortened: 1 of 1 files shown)


tanabi:
@wyld-sw If you do merge it, please take some time and test this as well for double verification.  I did a pretty big change here and I have this weird 'my head is telling me I should be confident but my gut is telling me I missed something' thing going on.

wyld-sw:
It seems fine to me. But you're right, we should probably make sure the code's not required by something that currently exists. The valgrind-bot combo is a great option here.

I'll go ahead and merge, but I'll let you know if anything seems off during additional testing.

Issue #451, opened 2019-08-14
Date Weirdness


tanabi:
In the latest fuzzball, this:

```
@muf "18:30:00 08/15/2019" convtime
```

produces the date in 2020 for some reason.  This:

```
@muf "18:30:00 08/15/19" convtime
```

seems like it works.  I thought we had fixed this date stuff before ?  Did we introduce some problem recently?

wyld-sw:
We added century support to years in {convtime} and the new CONVTIME, but that must have escaped my testing :(. I'll look into it, if you don't beat me to it!

wyld-sw:
I stand corrected; it's still a 2-digit year - and it seems to be reading the first 2 digits (20) as the year in your first example. There may be a way to prevent that - or actually recognize 4-digit years. We may need to code our own strptime for that?

wyld-sw:
Works well, thanks!

PR #452, opened 2019-08-15
Fixed string to time conversion, hopefully forevers.

tanabi:
@wyld-sw  Please give this some test pokes to make sure it looks good.  My test show it works acceptably.  There's probably a better/cleaner way to do this rather than my dumb pseudo-parsing but this is the best I could come up with on short notice.


tanabi:
This fixes #451 (just to link the issues)

Issue #457, opened 2019-09-15
ADDPENNIES primitive and GIVE command differ in behavior


dbenoy:
The ADDPENNIES primative will ensure that thing objects have positive values, but the GIVE command doesn't care.

Perhaps these behaviors should be harmonized. I think they should both prevent negative values, but a value of zero should be fine?

dbenoy:
A value of zero on an object is useful if you want some sort of process to be able to create objects in an unprivileged way, but you don't want users to be able to collect pennies when they recycle the object.

tanabi:
If I'm remembering right, recycle won't 'respect' negative pennies and will 'round' it up to 0 if an object has negative.  Though the MUF recycle command may operate differently as I am sure they use totally different sets of code.

Though I totally agree, these things should not only work the same, but they should share the same underlying code.

dbenoy:
Just gave it a try.

```
> @create test
Object test(#183) created.
> examine test
test(#183)  Owner: Mercury  Value: -1
Type: THING
Created:  Sun Sep 15 14:58:44 2019 EDT
Modified: Sun Sep 15 14:58:44 2019 EDT
Lastused: Sun Sep 15 14:58:44 2019 EDT
Usecount: 0
[ Use 'examine <object>=/' to list root properties. ]
Memory used: 227 bytes
Home: Cave of Awakening
Location: Mercury(#164PWQBC)
No actions attached.
> inventory
You are carrying:
test(#183)
You have 0 pennies.
> @recycle test
Thank you for recycling test (#183).
> inventory
You aren't carrying anything.
You have -1 pennies.
```

tanabi:
Well that's just nasty.  Yeah, we need to fix that :)  This is a pretty easy one.  Would you like to take a whack at it or you want @wyld-sw or myself to do it?

dbenoy:
So how would I go about it?

I've noticed GIVE lets wizards exceed tp_max_pennies when used on players feels like it makes sense.

But it also lets them set player penny amounts to negatives, and it has no checks at all for wizards when giving to thing objects.

```
> @tune max_object_endowment
(int)  max_object_endowment = 100
*done*
> give #187=10000000
You change the value of testobject to 10000000 pennies.
> give #187=-20000000
You change the value of testobject to -10000000 pennies.
> give me=-100
You take 100 pennies from Mercury.
Mercury takes 100 pennies from you!
> inventory
You are carrying:
testobject(#187)
You have -100 pennies.
```

[...]

PR #462, opened 2019-09-16
Recycling thing objects with negative value no longer costs you pennies

dbenoy:
Fixes issue #457

As discussed in that issue, it also brings ADDPENNIES behavior in line with the GIVE command.


wyld-sw:
Merged!

PR #464, opened 2019-09-17
Pennies related fixes to @chown and @link when dealing with unlinked exits

dbenoy:
Fixes two things:
- `@link unclaimedexit=room` when you don't have a BUILDER does not work, but it was charging you the pennies /before/ rejecting the attempt. I reversed this, checking first before charging.

- You have to pay tp_exit_cost pennies when you do `@link unclaimedexit=room`, but you could cheat and get around this by doing `@chown` unclaimed exit. I copied the logic for charging you pennies from `@link` into `@chown`.

IMHO though, I think that long term, the entire 'anybody can seize an unclaimed exit' thing should be eliminated from fbmuck entirely.
- I can't imagine any MUCK out there using this functionality. (Correct me if I'm wrong?)
- It can easily be replaced in a more robust and intuitive way by creating a MUF program to claim/create exits.
- It's a security risk, in case you accidentally leave an unclaimed exit in `#0`, for example. It could be claimed, renamed to 'say' and then given an MPI program to beam all your communication off to a malicious party, or otherwise wreak havoc.

### DIFF
--- a/src/create.c
+++ b/src/create.c
@@ -175,21 +175,21 @@ do_link(int descr, dbref player, const char *thing_name, const char *dest_name)
                     return;
                 }
             } else {
+                if (!Builder(player)) {
+                    notify(player, "Only authorized builders may seize exits.");
+                    return;
+                }
                 if (!payfor(player, tp_link_cost + tp_exit_cost)) {
                     notifyf(player, "It costs %d %s to link this exit.",
                             (tp_link_cost + tp_exit_cost),
                             (tp_link_cost + tp_exit_cost == 1) ? tp_penny : tp_pennies);
                     return;
-                } else if (!Builder(player)) {
-                    notify(player, "Only authorized builders may seize exits.");
-                    return;
-                } else {
-                    /* pay the owner for his loss */
-                    dbref owner = OWNER(thing);
-
-                    SETVALUE(owner, GETVALUE(owner) + tp_exit_cost);
-                    DBDIRTY(owner);
                 }
+                /* pay the owner for his loss */
+                dbref owner = OWNER(thing);
+
+                SETVALUE(owner, GETVALUE(owner) + tp_exit_cost);
+                DBDIRTY(owner);
             }
 
             /* link has been validated and paid for; do it */
--- a/src/set.c
+++ b/src/set.c
@@ -441,6 +441,23 @@ do_chown(int descr, dbref player, const char *name, const char *newowner)
         }
     }
 
+    /* handle costs */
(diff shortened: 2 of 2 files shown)


wyld-sw:
I'm unsure how many MUCKs still rely on unlinked exits to claim homes in public spaces. Removing it in favor of a purely-MUF solution *might* be too much for the first big release in years (or at least it would require an easy conversion process), but I do agree it can seem a bit wonky. 

This is a good conversation to have, but I'll be merging this in. Feel free to keep discussing it here.

dbenoy:
If you're worried, it could be a `@tune` instead (Or a config.h setting macro) that defaults to off, and if somebody upgrades and sees that it broke a MUCK feature they were using, they can just turn it back on.

But personally I think it might be good to put out some feelers to see if this functionality actually gets used. I expect the answer will be no.

tanabi:
In my nearly 30 years of MUCKing I've never seen the unlinked exit thing being used anywhere, except 'by accident'.  I will admit I have, at times, claimed unlinked exits just for fun on occasion :)

Something we can also consider is a deprecation warning.  We can put something in when someone claims an unlinked exit that says "Warning: this feature will be removed in the next version of Fuzzball" kind of thing so its not a SUPRISE!

wyld-sw:
Ok! I’m thinking about both a tune defaulted to off, AND a depreciation warning when it’s on. 

We should think about other features that need the deprecation treatment, if any :)

tanabi:
I don't think it makes sense to do both necessarily ... basically the deprecation warning saves you the work of making the tune variables, etc.  Because if the intent is to ultimately remove the feature, then we probably shouldn't add extra code to support it ya know?

So I think it should be one or the other, really.  But yeah if there's anything else we want to deprecate, this would be a good way to do it.

wyld-sw:
That’s fair. Deprecation it is!

wyld-sw:
Ok, 5756cac8c79e0fc82dc5615acc23162a31756aa6 adds the message to chown, link, and relink when acting on unlinked exits. Those last two should eventually use combined logic of course, as @tanabi suggested in the comments.

Please check if I screwed something up or forgot an important case. I seem to have had a rough time with "simple" fixes lately...

tanabi:
@wyld-sw I reviewed it and it looks good.

I think you can also claim an unlinked exit with @ chown ?  I'm not sure though, I'll try to test it here in a sec.

tanabi:
Yes, you can claim an unlinked exit with @ chown and I see it gets the message so no worries, change looks great.

PR #466, opened 2019-09-19
Fixed segfault on @link object=NIL

dbenoy:


wyld-sw:
Nice, thanks! That one was tough to initially track down. I now realize that the segfault didn't happen if teleport_to_player was yes.

tanabi:
@dbenoy Good job!  That one's been bothering us for awhile :)

PR #467, opened 2019-09-23
SETLINKS_ARRAY can now link to NIL

dbenoy:
SETLINKS was able to link to NIL, but not SETLINKS_ARRAY. Fixed.

<hr>

\<rant\>
This looks like a bug caused by large blocks of nearly-identical code being in multiple places in the source, and when a modification happens, forgetting to replicate the same change to other bits of the code.

It actually seems pretty common for big chunks of code with very similar logic to be just copied with small modifications all over the source. Might be worth going through and fixing that up.

One nuclear option to fix a big class of this code duplication problem (which I would actually be in favor of) would be to eliminate the built-in commands entirely and replace them all with MUF programs (with the exception of `@program` and probably some limited emergency commands in case you mess up `@action` or `@link`)

As a bonus the code would get a lot leaner, and wizards will have full freedom to change pretty much every command name or add whatever colors they want, etc, and if they wanted to nuke or permission-restrict or tighten a command due to some special security consideration specific to their MUCK (Like if they don't want you setting your own HOME for example), they'd have full freedom to do that.

I'd be happy to write MUF versions of built-in commands if you need a volunteer.

It might be something to save for fb8, though?

Edit: Maybe this suggestion is too ambitious...
\</rant\>

### DIFF
--- a/src/p_db.c
+++ b/src/p_db.c
@@ -2410,7 +2410,7 @@ prim_setlinks_array(PRIM_PROTOTYPE)
 	    array_data *val = array_getitem(arr, &idx);
 	    dbref where = val->data.objref;
 
-	    if ((where != HOME) && !valid_object(val)) {
+	    if ((where != HOME) && (where != NIL) && !valid_object(val)) {
 		CLEAR(&idx);
 		abort_interp("Invalid object. (2)");
 	    }
@@ -2422,6 +2422,9 @@ prim_setlinks_array(PRIM_PROTOTYPE)
 
 	    switch (Typeof(what)) {
 	    case TYPE_EXIT:
+                if (where == NIL)
+                    break;
+
 		switch (Typeof(where)) {
 		case TYPE_PLAYER:
 		case TYPE_ROOM:


wyld-sw:
Certainly we’ve got several tasks to consolidate logic.

I had a similar thought - separate to the logic issue - a while back, but got distracted with other fixes.

I think it might be worth looking into. What do you think, @tanabi? I agree that we should release fb7 first if we decide to do this.

tanabi:
MUF versions of a lot of commands already exist.  So doing this wouldn't be a stretch.  You have the MPI complication as well; are you going to implement all the MPI calls in MUF too?  In order to make this 'perfect' I think we would have to.

I think this is a reasonably good idea.  It would solve a lot of other issues, like the "giant switch statement" issue could potentially go away.

However, what would be the upgrade path for existing MUCKs?  Would we need to actually have "inline MUF code" in the .c files to support existing MUCKs?

There's a lot of "devil details" here.  But its worth exploring.

tanabi:
Something we may be able to do with MPI is "bridge" it to MUF primitives.  Many MPI calls have a 1 to 1 relationship with MUF calls, it's just they have a 'vector' of arguments instead of a stack.  The MPI functions could be turned into thin wrappers over MUF primitives under the hood.

Of course, using MUF to define MPI calls may be the ultimate, but I wouldn't want to fire off a MUF process for every MPI tag.  Optimally it would be some kind of library of MUF that maps functions to MPI prims.  I'm not sure how to do this without it having a lot of overhead though.  Maybe the overhead doesn't matter as much as I think it does, but, a MUCK like Tapestries or Furry where people are looking at things all the time and constantly slamming MPI might be rough.

Issue #459, opened 2019-09-15
Penny generation with NEWOBJECT primitive


dbenoy:
NEWOBJECT creates objects with a value of 1, rather than 0.

You can just create an object, and then recycle it to collect a penny.

This already can be exploited on numerous MUCKs, including my own, which have ways of ordinary users or low-level MUCKER bit users to create objects that they own from MUF code.

They could repeatedly create and recycle the object and endow themselves with as many pennies as they want.

tanabi:
Yeah, newobject (and probably newroom, etc.) doesn't deduct pennies to create but will set a value.  Most quota libraries that I've seen will deduct pennies and then set the penny value on the object according to the exorbitently complex way 'help @ create' describes (though I think @wyld-sw and I simplified that logic a good bit recently).  It would probably be a good idea to have these all calls not set an intrinsic value ... while that differs from how the commands work, the man pages don't hint that a value is being set and I think most programmers are not aware of this.

dbenoy:
Sadly there's no workaround for this either. Trying to eliminate the object's value with ADDPENNIES (in addition to requiring a wizbit on your script) gives you an error if you try to reduce the value below 1.

So no matter what, if a MUF program is creating an object for whatever reason, it's adding surprise pennies to your economy that most MUCK admins are probably not aware of, and may be confused about where they're coming from.

tanabi:
This is another really easy one.  Would you like to take a shot or you want @wyld-sw or myself to do it?  Trying to cultivate new FB contributors ya see, though I'm happy to do it if you're not comfortable.

dbenoy:
Sure I don't mind tackling it.

tanabi:
@dbenoy awesome!  Feel free to ping me if you have any questions, I am happy to help.

dbenoy:
Reposting this here. As mentioned in #461, Unfortunately, yes, newexit, newroom, setlink, and setlink_array do still have the same problem.

But unlike thing objects, they don't store their penny value on an object-by-object basis in the database :( So there really is no straightforward way to prevent @recycle from giving you free pennies for them, in the same way as for thing objects.

Not sure what can be done about it.

tanabi:
Yeah, I tend to forget that rooms and exits are special snowflakes.  I still think we should be able to make a code-path through so that MUF doesn't deduct pennies for those things, though its certainly a higher level of effort.

dbenoy:
[...]

PR #468, opened 2019-09-24
COPYOBJ was setting the penny value of _both_ thing objects to 1

dbenoy:
The COPYOBJ primitive duplicates the contents of the database object, but leaving a pointer to the previous properties tree in tact. Then it calls copy_prop which (I believe) makes a new property tree structure for the object and copies the previous one.

However, in the meantime, while the pointer for the property tree for both objects is still set to the original object, it does `SETVALUE(nu, 1)` which acts on the shared pointer (because the object's penny value is inside the property tree... I'm guessing it wasn't always this way) and sets the value of both objects to 1.

I don't know if I've addressed this correctly, though. Before merging this change... should I still be setting the object's penny value?

This pull request as written will copy the penny value exactly. Whatever the source object had the destination object will have. Theoretically, this could be part of the same class of problems I discussed in #459, though. Somebody with a low MUCKER level could use it combined with the RECYCLE primitive to generate pennies.

Maybe I should have it do a SETVALUE(nu, 0) after the copy_prop line? Or maybe this primitive should be changed to require a higher MUCKER level? Let me know if you want me to revise it.

Edit: Also the COPYOBJ primitive behavior is not in line with the `@clone` command. COPYOBJ will set the home to the player, `@clone` will leave the home alone. Maybe I should modify the THING_SET_HOME(nu, player); line as well, so it copies instead?

### DIFF
--- a/src/p_db.c
+++ b/src/p_db.c
@@ -572,6 +572,8 @@ prim_copyobj(PRIM_PROTOTYPE)
 	newobj = new_object();
 	*DBFETCH(newobj) = *DBFETCH(ref);
 	copyobj(player, ref, newobj);
+        if (mlev < 3)
+          SETVALUE(newobj, 0);
 	CLEAR(oper1);
 	PushObject(newobj);
     }
--- a/src/p_db.c
+++ b/src/p_db.c
@@ -41,7 +41,7 @@ copyobj(dbref player, dbref old, dbref nu)
     NAME(nu) = alloc_string(NAME(old));
     if (Typeof(old) == TYPE_THING) {
 	ALLOC_THING_SP(nu);
-	THING_SET_HOME(nu, player);
+	THING_SET_HOME(nu, THING_HOME(old));
     }
     newp->properties = copy_prop(old);
     newp->exits = NOTHING;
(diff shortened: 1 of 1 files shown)


dbenoy:
Maybe the copyobj primitive behavior should be brought in line with the `@clone` command?

If that's what you want, then this pull request is fine as is, but also I should remove the 'THING_SET_HOME(nu, player);' line. COPYOBJ sets the home to the player, but `@clone` does not.

I don't know what that ALLOC_THING_SP line does, gotta poke around a bit more.

tanabi:
copyobj has been the source of many bugs ... I think I've fixed one in the past as well.  No wonder, it sounds like there's some sloppy pointer work in there.

I believe it should copy the linkage (home) because that makes more sense.  I would say copy the penny value ... to my mind, a copy is a copy.  It shouldn't be a copy with caveats.  Or maybe it should only copy penny level on M3?  That may be the best of both worlds, it addresses your penny leak issue while staying flexible

dbenoy:
@tanabi Okay I've pushed those changes.

wyld-sw:
Should we call the copyobj function in both do_clone and prim_clone? After resolving the HOME differences and other potential weirdness. Maybe with a flag to determine whether to copy pennies or not (I don't think you can just check the player's mucker level)

dbenoy:
do_clone checks for how many pennies the object _would_ cost, and charges the user for it. So it's already safe from any penny shenanigans, so if we used the same routine in both, then in both cases the actual copy routine should not touch the pennies.

I figured I'd push these changes just to get rid of the bug for now, and maybe there should be a project entry or something for getting rid of duplicated logic.

PR #473, opened 2019-09-29
A @tune to prevent built-in commands (Except for wizards with !override)

dbenoy:
I'd like to add this tune because I want to use it on my own MUCK for security and cosmetic reasons, and because it seems like a good intermediate step toward some day removing built-in commands entirely.

I've also created a repository (https://github.com/dbenoy/mercury-muf) where I'm working on MUF replacements for built-in commands, among other things. I've created copies of the logic for the object creation commands and followed the engine source fairly closely.

For various reasons the commands won't have precicely the same message output, but they try to follow the behavior of the built in commands in a quirk-for-quirk way.

It also adds a few features like object quotas, but I intend, once these programs are fairly well tested in the wild, to create a separate repository with vanilla copies of the built-in commands with no enhancements.

### DIFF
--- a/docs/man.txt
+++ b/docs/man.txt
@@ -4442,6 +4442,7 @@ Parameters available:
  (str)  autolook_cmd              - Room entry look command
  (time) clean_interval            - Interval between memory/object cleanups
  (int)  cmd_log_threshold_msec    - Log commands that take longer than X millisecs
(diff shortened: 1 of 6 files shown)


tanabi:
Looks fine to me.  I've got no objection to this feature. @wyld-sw  ?

wyld-sw:
Makes sense to me too. Merged!

PR #489, opened 2020-01-12
First draft of p_connects documentation.

wyld-sw:


tanabi:
@wyld-sw one thing you didn't do is normalize the spacing :)  Looks like a mix of tab and space still.  I suppose I can take a pass at it if you don't want to do that part.

wyld-sw:
Yikes, thought I did that at the beginning. I'll take care of it.

wyld-sw:
Should be done now. Let me know if anything was missed.

PR #507, opened 2020-03-06
Documented p_misc.[ch]

wyld-sw:


src/p_misc.c
```diff
@@ -213,151 +393,282 @@ prim_fmttime(PRIM_PROTOTYPE)
...
     if (oper1->type != PROG_STRING)
-	abort_interp("Non-string argument.");
+        abort_interp("Non-string argument.");
+
     if (mlev < tp_userlog_mlev)
-	abort_interp("Permission Denied (mlev < tp_userlog_mlev)");
+        abort_interp("Permission Denied (mlev < tp_userlog_mlev)");
+
+    /**
+     * @TODO buf = DoNullInd(oper1->data.string);
```
tanabi: This is a pretty generic @TODO ... is the purpose to add the line that you have here?  Is there something else to do here?

src/p_misc.c
```diff
@@ -476,118 +810,195 @@ prim_fork(PRIM_PROTOTYPE)
...
+        /* tmp, ref */
+        rooms = exits = things = players = programs = garbage = 0;
+
+        for (dbref i = 0; i < db_top; i++) {
+            if (ref == NOTHING || OWNER(i) == ref) {
+                switch (Typeof(i)) {
+                    case TYPE_ROOM:
+                        rooms++;
+                        break;
+                    case TYPE_EXIT:
```
tanabi: To be technically correct (the best kind of correct) there should be a blank line between each 'case' block.  So like:

```
switch(Typeof(i)) {
    case TYPE_ROOM:
        rooms++;
        break;

    case TYPE_EXIT: /* and so forth */
```

src/p_misc.c
```diff
@@ -476,118 +810,195 @@ prim_fork(PRIM_PROTOTYPE)
...
-		garbage++;
-		break;
-	    }
-	}
+        if (ref == NOTHING || OWNER(i) == ref) {
+            switch (Typeof(i)) {
+                case TYPE_ROOM:
+                    rooms++;
+                    break;
+                case TYPE_EXIT:
```
tanabi: Same note as above re. switch spacing.

src/p_misc.c
```diff
@@ -821,115 +1429,217 @@ prim_setsysparm(PRIM_PROTOTYPE)
...
-    case TUNESET_BADVAL:
-        free(oldvalue);
-	abort_interp("Bad parameter value. (2)");
-    case TUNESET_DENIED:
-        free(oldvalue);
-	abort_interp("Permission denied. (1)");
-    default:
-        free(oldvalue);
-	break;
+        case TUNESET_SUCCESS:
```
tanabi: Same note as above re. case spacing.

src/p_misc.c
```diff
@@ -992,76 +1732,130 @@ prim_ext_name_okp(PRIM_PROTOTYPE)
...
-	for (ref = 0; buf[ref]; ref++)
-	    buf[ref] = tolower(buf[ref]);
-	if (!strcmp(buf, "e") || !strcmp(buf, "exit")) {
+        strcpyn(buf, sizeof(buf), DoNullInd(oper2->data.string));
+
+        for (ref = 0; buf[ref]; ref++)
+            buf[ref] = tolower(buf[ref]);
+
+        /**
+         * @TODO Maybe these should use case-insenstive?
```
tanabi: Agreed.  I'd take it a step further and strncmp them based on the length of the input string.

So for example, "ex" would match "exit".  But doing case insensitive is trivial so we should definitely do that :D

src/p_misc.c
wyld-sw: Changed to say that it may be able to replace the "if" construct entirely.

src/p_misc.c
wyld-sw: Done.

src/p_misc.c
wyld-sw: Done.

src/p_misc.c
wyld-sw: Done.

src/p_misc.c
wyld-sw: Might need a special case for "p" - does it match "player" or "program"? Also, should "m" and "mu" also be aliases for "muf"?

I'm sure it's doable.

Issue #510, opened 2020-04-05
Lack of MCP argument checking leads to crashes


revarbat:
If the `mcp_mesg_arg_getline()` function is called for a non-existing argument on an MCP message, it returns NULL.  Not all places that use this function are properly verifying that NULL wasn't returned.  This can cause server crashes without a user even logging in.  This is being actively exploited in the wild as a denial-of-service attack on popular FBMUCKs.

I have verified that this bug still exists in FB7, in the MCP simple-edit package.  In my (late FB6) copy of FurryMuck's server, I had to apply fixes in mcp.c, mcppkgs.c, help.c, and p_mcp.c

cyveris:
Is disabling MCP support during recompile sufficient to guard against this as a short-term measure?

tanabi:
I'm on it, I'll have a patch out shortly for FB7.

wyld-sw:
Thank you, @tanabi!

revarbat:
Disabling MCP support should work.

tanabi:
I've got a fix, I'm working on a test case.  I've asked @revarbat over email for whatever verified crash script we can use.

tanabi:
I've created my own test case that crashes the MUCK so I'll have a patch here soon.

revarbat:
I don’t have a script to reproduce. Bug was found by gdb analysis of a crash core file.
Connected to the server, did NOT log in to a user, negotiated MCP packages including simpleedit, then sent a simpleedit set message that lacked a reference argument.

tanabi:
@revarbat I have a script if you want it :)  I've found another problem where notify can get called when a player isn't logged in yet and that also causes a crash.

revarbat:
I would be interested in info on that latter bug.

tanabi:
@revarbat Please check out my PR : https://github.com/fuzzball-muck/fuzzball/pull/511   Basically there's an error handling call in mcppkg.c that has an improper notify.  I took SendText from mcp.c and turned it from a static to a proper function and just used that instead.

tanabi:
This is probably fixed.  I am going to do some more testing and leave this open for a day or two until I feel reasonably certain.

revarbat:
There are places where the result of MCP_PLAYER() is not being validated.  Most are okay, as `controls()` checks against -1, but there's at least one place a `Wizard()` test will cause a crash if the connection isn't logged in.

tanabi:
[...]

PR #511, opened 2020-04-05
Mcp crash fix

tanabi:
@revarbat  seems to be finding a lot more of these bugs than I am -- there are some calling patterns like this:

```
lines = mcp_mesg_arg_linecount(msg, "data");

for (int i = 0; i < lines; i++) {
    /* Any mcp_mesg_arg_getline for "data" in this loop should be safe, right? */
}
```

So Revar may be adding checks in there which I am not sure are necessary.  However, I would appreciate additional sets of eyes on this.

Fix for: https://github.com/fuzzball-muck/fuzzball/issues/510

### DIFF
--- a/include/mcp.h
+++ b/include/mcp.h
@@ -388,6 +388,28 @@ void mcp_package_register(const char *pkgname, McpVer minver, McpVer maxver,
                           ContextCleanup_CB cleanup);
 
 /**
+ * Send some text to a given MCP Frame
+ *
+ * The send is not immediate, it is queued to the descriptor with queue_write
+ *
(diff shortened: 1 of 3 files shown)


revarbat:
Actually, 'm looking at some of the uses of MCPFRAME_PLAYER, and the result might not be validly checked for being -1 in various places.

revarbat:
Hmm.  Given that `controls()` tests against -1, we may be okay.

revarbat:
No, the Wizard(player) check on the sysparm branch of simpleedit-set  can cause a crash for an un-logged in connection.

PR #512, opened 2020-04-05
Corrected some places where segfaults can occur if MCP is called via the login screen instead of being logged in.

tanabi:
For https://github.com/fuzzball-muck/fuzzball/issues/510


tanabi:
DO NOT MERGE -- should be ObjExists, not OkRef, whoops

PR #528, opened 2020-05-03
Addressed format-truncation compiler warnings, etc.

wyld-sw:


tanabi:
@wyld-sw This broke the windows build fyi ...

```
sanity.c
.\src\sanity.c(1121): error C2057: expected constant expression
.\src\sanity.c(1121): error C2466: cannot allocate an array of constant size 0
.\src\sanity.c(1121): error C2133: 'player_name': unknown size
.\src\sanity.c(1135): warning C4034: sizeof returns 0
.\src\sanity.c(1314): error C2057: expected constant expression
.\src\sanity.c(1314): error C2466: cannot allocate an array of constant size 0
.\src\sanity.c(1314): error C2133: 'name': unknown size
.\src\sanity.c(1322): warning C4034: sizeof returns 0
```

src/sanity.c
```diff
@@ -1118,7 +1118,7 @@ rand_password(void)
 static void
 create_lostandfound(dbref * player, dbref * room)
 {
-    char *player_name = "lost+found";
+    char player_name[tp_player_name_limit+1];
```
tanabi, line 1121: This is probably what Windows is complaining about.  Most compilers can't do dynamic size allocations of heap variables like this, I don't think.  You have two options -- use a static size, or do like:

```
char * player_name = (char*) malloc(tp_player_name_limit+1);
```
and don't forget to free on returns.  I think it is probably important to address this as it could break in random other compilers.

src/sanity.c
```diff
@@ -1309,33 +1309,13 @@ find_misplaced_objects(void)
                 case TYPE_GARBAGE:
                     NAME(loop) = "<garbage>";
                     break;
+
                 case TYPE_PLAYER: {
-                    char *name = "Unnamed";
+                    char name[tp_player_name_limit+1];
```
tanabi, line 1314: Same problem as above.

dinoex:
Please not the difference here:

    char *player_name = "lost+found";

allocates a pointer on the stack, that points in the data segment with the constant string. There is no buffer allocated.

Better:
    const char *player_name = "lost+found";

This allocates a buffer on the stack:
    char player_name[tp_player_name_limit+1];

tanabi:
> 
> Allocates a pointer on the stack, that points in the data segment with the constant string. There is no buffer allocated.
> 

The problem being solved here is that these 'constant' strings are being overwritten in the code under it ... So it needs to be changed to a properly sized memory buffer with the correct initial string.  This is also a really infrequent call (I'm pretty sure in the history of fuzzball it has never been used, because it would segfault if it ever was used :) ) so the malloc approach is best.  Malloc the buffer, strcpy the initial string, and make sure to free it later.  It's just not a good use case for a stack initialized string.

PR #531, opened 2020-05-10
Fix DESCR_ARRAY crash

Loonesbury:
An 'else' in prim_descr_array() was removed in b09e0064, causing segfaults on '#-1 DESCR_ARRAY'.


wyld-sw:
Thanks for the catch!

PR #532, opened 2020-05-10
Added timequeue documentation

tanabi:
LAST ONE OMG!

### DIFF
--- a/src/timequeue.c
+++ b/src/timequeue.c
@@ -30,43 +30,71 @@
 #include "timequeue.h"
 #include "tune.h"
 
-#define TQ_MUF_TYP 0
-#define TQ_MPI_TYP 1
+/*
+ * TODO: Figure out what "TQ_MUF_TREAD" is.  As far as I can tell,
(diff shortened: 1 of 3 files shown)


src/timequeue.c
```diff
@@ -30,43 +30,71 @@
 #include "timequeue.h"
 #include "tune.h"
 
-#define TQ_MUF_TYP 0
-#define TQ_MPI_TYP 1
+/*
+ * TODO: Figure out what "TQ_MUF_TREAD" is.  As far as I can tell,
+ *       we never set it.  If it is unused, we should remove the reference
+ *       to it -- it'll simplify a lot of if statements.
+ */
```
wyld-sw, line 37: Guessing here, but “timed read”? TREAD is an in-server define,  it it may have been different at one time. Agreed we should clean it up if unused.

tanabi: It *is* used I think ... maybe ... It's a very twisty code path.  :)

PR #539, opened 2020-05-18
Some automated tests

woggle:
This would add support for automated testing, with some basic tests that for some commands (in YAML files named after where the commands are implemented). It also makes .travis.yml run the tests.

The testing code is written in Python and requires Python 3.5 (and will not AFAIK work on Windows). I welcome comments on better structuring it.

Each test works by loading a minimal DB in a temporary directory and running the server with -console, sending certain input and looking at the output. To avoid output being truncated, the tests wait for certain output to appear before sending other input. This prevents, for example, '@shutdown' from taking place before queued output from a previous command is written. The tests use the 'pose' command to create markers in the output to know when a certain set of commands is complete.


tests/test_util.py
```diff
@@ -0,0 +1,180 @@
+import asyncio
```
tanabi, line 1: Some general comments on this whole file ...

This is a really cool system you've built but I'd like to see some documentation on it.  A little header block explaining how this is expected to be run, any notes about how to build new tests, would be great.

Additionally, Pythonic doc-blocks for the functions would be good.  I know its a lot to ask for something that's done on your free time, but I just spent a year of my life documenting code so asking for a little bit of docs from others seems fair :D 

Pythonic docblocks look something like:
```
def myfunction(arg1, arg2, arg3):
    """This is a brief description of the code

    Args:
        arg1: str that does this
        arg2: int that does that
        arg3: dict that does whatever and has these elements

    Returns:
        A whole sandwich
    """
```

'self' doesn't need docs when passed as a parameter because that's a language feature.  If you want to do python3 type hinting for arguments that is also great but not required.

Thanks for cooperation :)

woggle:
I think the compilation errors were a bad merge; should be in a sane state now.

tanabi:
Approved and merged!  Thanks!

Issue #537, opened 2020-05-17
We need a DUMP prim!


cyveris:
Currently, the only configurable for database dumps is how often. We'd like to be able to configure dumps to happen at a specific server time, e.g. hourly **on the hour**, similar to a cron job.

wyld-sw:
How about an option (`@tune cron_program`?) that specifies a MUF to run every so often, and expose dumping as a wizard primitive?

Just an idea. Let me know if it's not feasible.

cyveris:
That sounds perfect, really. The purpose behind the request is ensuring internal consistency for regular backups/rsync jobs. Don’t want to accidentally get into a timing where rsync is occurring during a dump.

cyveris:
That could be perfect. The purpose behind the request is ensuring internal consistency for regular backups/rsync jobs. Don’t want to accidentally get into a timing where rsync is occurring during a dump.

The question is how often the `cron_program` would run, and how we'd ensure that it runs at the minute desired.

wyld-sw:
What are your thoughts on that? I was thinking it could run every minute as you might expect - which doesn't seem like it would be an issue normally. Could be completely wrong though.

cyveris:
That's as often as anything in a crontab or `cron.d` file would be executed. I don't see an issue there.

wyld-sw:
Could also have a `@tune cron_interval` defaulting to 1m if there's a need to change it for performance reasons.

cyveris:
Love it!

tanabi:
Here's a couple more thoughts on this --

A 'crond' could actually be very easily programmed in MUF.  There's no particular reason to server-side this.  If you'd like me to write this, I could probably do it pretty closely.

If we're going to bake it into the server, I'd want to make it more flexible than to run a single MUF at some interval ... to me that seems like barely better than writing crond in MUF :)

tanabi:
If crond is written in MUF, we can make it basically work like real cron.  Could even use the same interval format and a proplist of entries.  Such as:

```
5 * * * * \#12345 This Goes On Stack
```

would call \#12345 every hour at the 5 minute mark.

cyveris:
We already have that. There are two things we need: The ability to disable database dumping via tune, and a wizard-level prim that allows a dump to be performed in MUF.

tanabi:
> We already have that. There are two things we need: The ability to disable database dumping via tune, and a wizard-level prim that allows a dump to be performed in MUF.

[...]

PR #551, opened 2020-05-19
Added stuff in support of DUMP primitive #537

tanabi:
Added DUMP prim, example MUF and Python.


wyld-sw:
Looks good! Thank you.

PR #560, opened 2020-06-01
Moved or reworked some defines.

wyld-sw:


wyld-sw:
Let me know if any of these repurposed config variables would be better as something else.

PR #562, opened 2020-07-09
Promote some variables to const, reduce instances of casting.

wyld-sw:
This is most of a large patch from Dinoex, focusing on reducing the amount of casting required while promoting some variables to const pointers (and a few minor cleanups). I'm still going through the remaining bits. I've tested this, and it doesn't seem to cause any trouble (or even seem like it SHOULD), but please feel free to scrutinize! :)


tanabi:
You know me, I love to scrutinize!  Taking a look.

PR #582, opened 2020-12-08
Introduced starterdb, a streamlined advancedb.

wyld-sw:
Removed basedb.
Removed note that "news index" is a valid command by default.
Added new default welcome screen.


tanabi:
This has merge conflicts because I edited advancedb/basedb to add that new tp parameter.  If you need help resolving the conflicts, let me know, but my changes to those files can be just discarded.

Issue #542, opened 2020-05-18
Documentation Artifact


tanabi:
It would be cool if we could auto-build documentation and publish it somewhere.  The command to build the docs is simply:

```
doxygen Doxyfile
```

it needs doxygen and graphviz packages installed and it makes an api-docs directory which can be zipped into a release file or published on the web.  Would be trivial to build docs with a Docker container or maybe there's some build option the same way the windows binaries are built.

I'm not too sure what the options here are, but I think its a cool goal.

tanabi:
I suppose I'll look into this one :)

tanabi:
This artifact is now part of the appveyor build.

PR #594, opened 2021-02-11
Make doxygen artifact

tanabi:
This is for issue #542 -- I am honestly not sure if this will work!

https://help.appveyor.com/discussions/problems/863-doxygen-installed-via-chocolatey-not-found

suggests it will, but I'm not sure if 'choco' is a command that comes with appveyor or not and my searches on google are inconclusive.  The only way I can think it do it is simply to try it.


tanabi:
@wyld-sw This works!  Appveyer is now making a doc artifact in addition to the windows zip artifact.

tanabi:
This should be pretty non-contraversial so I'm going to boop it into existance.  Due to how our build system works, it is already building the artifact :)  @wyld-sw if you think this should work differently then let me know ... we can always tweak it later!

PR #644, opened 2021-07-07
#607 Add SMTP Support to the MUCK - SMTP_SEND

### LINKED ISSUE #607: Add SMTP Support to MUCK
I'm posting this issue for feature visibility more than anything -- I have actually already programmed this!  I just need to test it thoroughly.

For Hope Island, I wanted to make the registration workflow work like this:

* User signs up on website
* Wizards on MUCK get a notification with the registration info, user is put on queue
* Wizards approve or deny character
* Email is sent to user.

The problem is "email is sent to user" ... there's no way for a MUCK to trigger an external action.  Either I have something poll the MUCK, I enable the MUCK to run local commands (Didn't want to do this for security reasons) or I add SMTP support.

It turns out SMTP support can be added with a single .h and .c file.  https://github.com/somnisoft/smtp-client is available with a creative commons license and is designed to just be included in your project.

A problem that may come up during testing is it may need to be forked into its own process.  I will try to fix this by making the timeouts more aggressive because I want the caller to get an error if there is a failure rather than assume success and have emails going nowhere.

I've implemented this with some `@tune` params and the new SMTP_SEND MUF primitive.  It could be used for all kinds of things -- event notifications, monitoring, triggering external processes, etc. - so I think it is a valuable addition.  And if you don't want it, simply don't configure it and it will be dormant.

tanabi:
Resolves #607 

This is pretty experimental still, but, it seems to work adequately.  Most notably is, it can lock the MUCK up until it times out.  That sucks, but, it's what I've got for right now.  I have already added a ticket to improve it.

@wyld-sw Please give this a careful look and make sure I didn't miss anything ... adding tune parameters AND a MUF call, there's a lot of "stuff" that has to be documented and added when doing that.  Easy to forget something :)


tanabi:
@wyld-sw  are we good to merge this you think?

tanabi:
Note from Wyld:

```
Wyld pages, "soorry, just got home from a long day. It looks good, though I'd 
probably suggest to just have the tunes have a blank "module". Having 
something there is just a way (currently) to say "these tunes are only enabled 
if the associated #define is turned on."" to you.
Wyld pages, "right now they show up as [inacive] in the tune list online, 
which is just cosmetic really." to you.
```

I'll update the doxygen documentation to explain that as well

PR #649, opened 2021-07-17
Clean up docker container implementation

skylerbunny:
- Make builds independant of the OS the container runs on
- Leverage tinit to allow better process handling
- Support non-default kill signals for the fbmuck process
- Update documentation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,36 @@
 FROM ubuntu:20.04
+RUN apt update && apt dist-upgrade -y
+RUN apt-get install -y build-essential \
+      libpcre3-dev libssl-dev git autoconf \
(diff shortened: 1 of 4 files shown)


tanabi (approved):
The only thing I would want to change is to make this build source from the local check out rather than cloning master -- I actually use the docker-compose process to test my local code changes :)  This, however, is a change I can do myself and overall this is a huge improvement to the overall process, so thanks for providing this.

skylerbunny:
You're right; when I set this up I was thinking about in terms of the Dockerfile being able to be run without the repository at all, but your proposed fix would make it flexible for "an end user who just wants to build and run a branch", and for" a developer who wants to test local changes one at a time".

PR #650, opened 2021-07-25
Fixed docker to build based on local source files

tanabi:
Fixes @skylerbunny 's PR using local files to build instead of pulling it down from GIT.  The build is actually still pretty fast so even though I feel like I might be able to make this more efficient (using volumes or whatnot) I don't think its worth the trouble.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,10 +3,13 @@ RUN apt update && apt dist-upgrade -y
 RUN apt-get install -y build-essential \
       libpcre3-dev libssl-dev git autoconf \
       automake autoconf-archive
-RUN git clone https://github.com/fuzzball-muck/fuzzball.git && \
-    cd fuzzball && git pull origin master && \
+COPY . fuzzball/
+RUN cd fuzzball && \
     ./configure --with-ssl --prefix /root/scratch && make clean && \
-    make && make install
+    make && make install && cd docs && \
+    ../src/prochelp ../src/mpihelp.raw mpihelp.txt mpihelp.html && \
+    ../src/prochelp ../src/mufman.raw mufman.txt mufman.html && \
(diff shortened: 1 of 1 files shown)


skylerbunny:
I think figured out why the addition of your lines to build the help files confused me as 'Why is that necessary?' In part I didn't realize that the help files are built from `.raw` files at build time. Now that I understand that...

First, an easy error to fix, and the Makefile does this right: it should be `prochelp ../src/muckhelp.raw help.txt muckhelp.html` (note `help.txt`, not `muckhelp.txt`). However, I suspect that these lines still shouldn't be needed in the Dockerfile, because I believe they do already run, and what this reveals is a different issue. Read on...

--

Playing around in the container, it looks like running `make install` implicitly runs `make help` before installing into your `${PREFIX}`. What `prochelp` builds, and where those files are copied, though, is `${PREFIX}/share/doc/fbmuck`. This is good implementation of global files, and would in fact have been copied from the build container to the run container.

BUT...what an individual `fbmuck` process seems to want these files to live in, however, is `<Your-Muck-Data-Directory>/data`. That's why I made that copy step later in the Dockerfile that's specific to the basedb. I now realize THAT should be unnecessary too...

I suspect to make this seamless (and this would be true for both Docker and non-Docker implementations of Fuzzball), the fbmuck process itself should look for the help files _where they're installed globally_: `${PREFIX}/share/doc/fbmuck` and not inside some given MUCK's data directory.

skylerbunny:
Hmm. To fully elaborate, now that I've taken a few minutes to think on it, maybe I'd amend this to (realizing that 'But I want server specific help files to still be possible if someone wants to copy them there'), maybe the behavior of the `fbmuck` process, for the help files, wants to be:

Check `<Your-Muck-Data-Directory/data`. If not found...
Check `${PREFIX}/share/doc/fbmuck` If not found...
Fail with an error.

Issue #566, opened 2020-07-19
Consider making descriptors and connection numbers synonymous


wyld-sw:
DESRCON and CONDESCR would be no-ops. Look into this further to see if this is feasible.

wyld-sw:
I think this makes a lot of sense (was @tanabi's idea). I'm trying to think of when someone would need to use the connection numbers directly. Sure, you could do `1 contime`, but is there a use case for something like that? Is there something (besides DESRCON) that returns a connection number?

This might be another case of introducing something new (descriptors in MUF) but wanting to still support the previous method for a while. I could be missing something, for sure.

If we did this, maybe:
- Alias CONBOOT, CONDBREF, CONDESCR, CONHOST, CONIDLE, CONNOTIFY, CONTIME, and CONUSER to their DESCR* counterparts.
- Keep CONCOUNT (potentially rename/alias to DESCRCOUNT).
- Make CONDESCR and DESCRCON no-ops.
- Test with common MUFs that use CON* primitives.

wyld-sw:
Oh ... "looping from 1 to CONCOUNT" might be a use case for the connection numbers. If that's the case, do we alias things like CONIDLE to CONDESCR DESCRIDLE, and "hide" those CON* primitives from normal use?

Hmm.

tanabi:
> Oh ... "looping from 1 to CONCOUNT" might be a use case for the connection numbers. If that's the case, do we create in-server defines for things like CONIDLE, to be CONDESCR DESCRIDLE?
> 
> Hmm.

This DOES potentially make a complication ... on Hope Island there's only 1 program that does such a loop (cmd-3who.muf) so I think this is a lot more common with older MUFs.  It may just be a breaking change and perhaps we fix MUFs in fbmuf github.  We could make a "generator" style call (kind of like NEXT .... NEXTDESCR ? ) so that instead of doing a 1 ... CONCOUNT loop, you can loop on NEXTDESCR for a similar experience.  That would be a very simple change and easy for folks to port to.

tanabi:
Looks like NEXTDESCR already exists :)

wyld-sw:
#645 maps CON* in-server defines to DESCR* counterparts. Please review.

This will break existing MUF code that relies on concount loops, but the fix for that situation is straightforward (see change to starterdb's 85.m). At the very least we'll need a communication plan ( and possibly links to updated program(s) )before this goes live.

tanabi:
[...]

PR #655, opened 2021-08-08
Final version of condescr work

tanabi:
This is for #566 

I am pretty sure this works now.  I've been through basically everything and it seems like I've worked out all the kinks.  Found one last serious bug with it and corrected another users displaying in the wrong order issue (This time with 3who)

The only thing I haven't done is gone through the fuzzball-muf repo and make sure all that stuff still works or is modified accordingly.  @wyld-sw have you looked at that at all yet?  It probably needs to be done as well, maybe even before this is merged?  Not sure how much we care.

@skylerbunny @jcoffm When this merges, this may break WHO / wd / 3who on your MUCKs.  Kind of a final warning about this potentially breaking change :)

Oh one more thing @wyld-sw I am unable to delete the 'condescr' branch from the fuzzball main project.  You'll have to clean that up.  I think the protected branch settings are preventing it from being cleaned up.  Let me know if you need help sorting that, its a little fidgety.


wyld-sw:
I’ve been looking at fuzzball-muf a bit, and have some changes set to go (probably the first of a few) after this was approved - but it can happen anytime, really.

PR #657, opened 2021-09-22
More condescr fixes

Loonesbury:
This fixes bugs in DESCR_ARRAY, ONLINE, and ONLINE_ARRAY that weren't caught in #655. To demonstrate:
```forth
: main
    #-1 descriptors debug_line popn
    #-1 descr_array debug_line pop
    online debug_line popn
    online_array debug_line pop
;
```

It also fixes ONLINE and ONLINE_ARRAY returning multiple dbrefs for players that have more than one descriptor connected and logged in. This was inadvertently introduced when migrating from the "connection" list to the descriptor list.

Regarding my comments in d1de0a9, I'm honestly unsure whether to replicate the original undocumented behavior. I don't know of any code that uses it besides one program from Nameless MUCK, which sorts the connection list itself anyway according to user input. If you'd like the old behavior back, just say the word.

### DIFF
--- a/src/p_connects.c
+++ b/src/p_connects.c
@@ -733,17 +733,27 @@ prim_descr_array(PRIM_PROTOTYPE)
     temp2.type = PROG_INTEGER;
 
     if (ref == NOTHING) {
-        d = descriptor_list;
+        result = 0;
+
+        for (d = descriptor_list; d; d = d->next) {
+            if (d->connected)
+                result++;
+        }
+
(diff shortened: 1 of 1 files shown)


tanabi (approved):
Makes sense to me.

tanabi:
The appveyor build failure appears unrelated to this PR, I will make a separate issue for it.

tanabi:
Thanks for the help :)  This was a pretty big change and I was sure something slipped through the cracks even though we tested it as best we could.

PR #659, opened 2021-09-28
Fix possible uninitialized var in DESCRNOTIFY

Loonesbury:
Additionally, it didn't always check whether the descriptor was valid. Not
strictly a huge defect, but you shouldn't be able to run `-1 "" DESCRNOTIFY`
and *not* get an error.

### DIFF
--- a/src/p_connects.c
+++ b/src/p_connects.c
@@ -626,8 +626,15 @@ prim_descr_notify(PRIM_PROTOTYPE)
     if (oper2->type != PROG_STRING)
         abort_interp("Argument not an string. (2)");
 
-    if (oper2->data.string)
+    if (oper2->data.string) {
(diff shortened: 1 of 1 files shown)


tanabi (approved):

tanabi:
Thank you!

Loonesbury:
I wasn't entirely certain that this wasn't intentional - it'd been there for a long time...

tanabi:
I'd actually say its likely that nobody cared about the condition rather than it being intentional.  Fuzzball has a lot of odd conditions where if you do something unexpected it just sorta silently does nothing (like in this case) rather than give errors.

I try to fix these when I notice them, so in my book, this is a more appropriate way to handle it than to let it silently carry on and causing a potentially nonsense error further down in the code.

PR #663, opened 2022-01-26
Tiny fixes

aidanPB:
Two itty bitty changes:
1. Take care of the TODO in ``prim_array_get_propvals``, which the doc comment claimed was already implemented; and
2. Pull the check capping the returned array in ``prim_array_get_proplist`` out of the loop.

### DIFF
--- a/src/p_array.c
+++ b/src/p_array.c
@@ -1866,27 +1866,7 @@ prim_array_get_propvals(PRIM_PROTOTYPE)
                 }
 
                 if (goodflag) {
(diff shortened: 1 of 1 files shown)


wyld-sw:
Seems like a good idea. The default value of max_propfetch is double what was hardcoded, but I don't think I see a problem with that.

tanabi:
@aidanPB @wyld-sw looks good to me, thanks for doing this work AidenPB!

PR #665, opened 2022-02-08
wiz.c changes

wyld-sw:
* Changed some [un]bless logic: The strict_god_priv check happens outside the loop, and unbless includes that check now, and both use buffers spaced to avoid warnings for the response.
* Changed usage output to expect larger number ranges.
* Refactored profiling command set into one command - `@tops`:
- This command now has an optional parameter [muf|mpi] that can switch to the corresponding logic.
- This is mostly moving logic around, so any profound updates (stop reading the entire database, stop sorting linked lists, etc.) have not been made.
- I believe there were some (more) issues in the original code as a result of copying into the other two commands. The last line of output in particular can show over 100% idle time and the "profile length" only seems to be reset when you "reset" all objects.

Please test. I don't think I made things worse. Also feel free to modify if needed.

### DIFF
--- a/src/wiz.c
+++ b/src/wiz.c
@@ -334,20 +334,6 @@ blessprops_wildcard(dbref player, dbref thing, const char *dir,
     int i, cnt = 0;
     int recurse = 0;
 
-    /* @TODO: This check is done by the caller methods so this is always
-     *        redundant.  Worse: this is a recursive call, so we're punching
-     *        this check over and over for no reason.  Granted, blessing
-     *        doesn't come up that much :)  But still!  Remove this check
-     *        and add a note to the docblock that the caller is responsible
-     *        for permission checks.
-     */
-#ifdef GOD_PRIV
-    if (tp_strict_god_priv && !God(player) && God(OWNER(thing))) {
-        notify(player, "Only God may touch what is God's.");
-        return 0;
-    }
-#endif
-
     strcpyn(wld, sizeof(wld), wild);
     i = strlen(wld);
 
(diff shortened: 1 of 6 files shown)


src/wiz.c
```diff
@@ -313,6 +313,8 @@ do_teleport(int descr, dbref player, const char *arg1, const char *arg2)
  * Two kinds of wildcards are understood; * and **.  ** is a recursive
  * wildcard, * is just a 'local' match.
  *
+ * The caller is responsible for permission checks.
```
tanabi, line 316: Is this string update also in the header file?

src/wiz.c
```diff
@@ -1069,266 +1051,75 @@ do_usage(dbref player)
...
- * 'struct profnode' linked list.  For large numbers of 'arg1', this is
- * really inefficient since its sorting a linked list.
- *
- * This does not do any permission checking.
- *
- * @param player the player doing the call
- * @param arg1 Either a string containing a number, the word "reset", or ""
+ * Adds a 'struct profnode' to a given ordered linked list, if the size
+ * limit has not been reached.
+ *
```
tanabi, line 1056: I think `@private` has to be added to static functions

tanabi (changes requested):
Couple of minor documentation details, but overall this looks good.  Not sure when I'll have a chance to really put it through the paces though, but I think this is relatively safe.  Confirm documentation bits and then I think it is okay to merge.

src/wiz.c
wyld-sw: That's a valid concern, though I don't think there's a prototype anywhere.

src/wiz.c
wyld-sw: Thanks! New commit incoming.

src/wiz.c
tanabi: You're right, I missed that this is a static due to how git hub clips the file.  Derp!

tanabi (approved):

PR #668, opened 2022-11-24
Fix MUF error messages not logging to .debug

Loonesbury:
Fixes a bug introduced way back in e35d6543 -- `buf` was no longer being written to, so `.debug/lasterr` wasn't getting written properly.


tanabi:
Looks good to me!  Thanks!

PR #669, opened 2022-11-27
Revert to old version format

Loonesbury:
Unsure whether to make a this a pull request, an issue ticket, or a comment on the original commit, especially since I'm 2 years after the fact. Feel free to critique.

The new version number format, as of 1ee836a4, sorts _before_ previous version numbers. It also breaks the promise in `man __version` that `__version` will always start with `Muck2.2fb`. Granted if the old license isn't in effect anymore the version number doesn't _have_ to start with anything, but given that it breaks every version comparison ever written, I think this is a pretty major change.

### DIFF
--- a/docs/help.txt
+++ b/docs/help.txt
@@ -1,4 +1,4 @@
-                     MUCK Reference Manual for fb7.0
+                 MUCK Reference Manual for Muck2.2fb7.00
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/docs/man.txt
+++ b/docs/man.txt
@@ -1,4 +1,4 @@
-                     MUF Reference Manual for fb7.0
+                 MUF Reference Manual for Muck2.2fb7.00
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/docs/mpihelp.txt
+++ b/docs/mpihelp.txt
@@ -1,4 +1,4 @@
-                     MPI Reference Manual for fb7.0
+                 MPI Reference Manual for Muck2.2fb7.00
                    by Revar Desmera <revar@belfry.com>
 
 You may get a listing of topics that you can get help on, either sorted
--- a/include/config.h
+++ b/include/config.h
@@ -19,7 +19,7 @@
 /**
  * Server version number
  */
-#define VERSION "fb7.0"
+#define VERSION "Muck2.2fb7.00"
 
 /************************************************************************
    Administrative Options


tanabi:
@wyld-sw Hey Wyld, what do you think of this one?  I don't have a strong opinion ... I don't really like the extra unnecessary numbers, but I have to agree that @Loonesbury has a point that it could break version comparison theoretically.

That said, the old version number was so jenky it doesn't really follow modern conventions much, so I think a 'stock' version processing library (if such a thing exists .. might be something to look into) would probably have a harder time with the old version number.

@Loonesbury was this issue causing you pain in some way?  i.e. scripting issue, etc.

Loonesbury:
This came up when I inherited a bunch of legacy programs from FB5 that rely on `$ifdef __version<Muck2.2fb...` to behave properly.

On the one hand, it's not the end of the world to go through and remove those, especially since some of them are old enough that they actually predate `PARSEPROP`:
```
$ifndef __version<Muck2.2fb5.0
  "look" 0 parseprop
$else
  getpropstr
$endif
```
```
$ifdef __VERSION<MUCK2.2fb4.5
  trig not if #0 "_uptime" "" systime addprop exit then
  #0 "_uptime" getpropval
$else
  #0 "_sys/startuptime" getpropval
$endif
```
On the other hand, this effectively severed backwards compatibility with FB5 and FB6 -- $ifdef can only do raw string comparisons, and adding complex semantic version handling to the MUF preprocessor because of a format change is a less-than-optimal idea -- so I wanted to at least raise the issue for discussion.

Loonesbury:
To amend myself, it doesn't make backwards compatibility impossible, just really clumsy -- this is my best attempt at some typical version checks:
```
$ifndef __version<Muck2.2
    ( new code )
$else
$ifndef __version>Muck2.2fb6.14
    ( duplicate new code )
$else
    ( old code )
$endif
$endif
```

tanabi:
Interesting case.  So you're maintaining these programs to preserve FB 5 compatibility?  ... Is there actually a MUCK out there still running FB 5 or are you doing this for 'good stewardship'?  I ask because FB5 is like 20 or 30 years out of date now, there 'should be' nobody running it anymore.

That said, overall I think you're right -- the statement in the manual page that says the number will always start with Muck2.2fb is kind of our 'programmer contract' and we violated it here by accident.  I'd still like @wyld-sw to chime in (he's probably doing family stuff over the holiday) before approving this, but I'm in favor of correcting the version number.

Loonesbury:
> Interesting case. So you're maintaining these programs to preserve FB 5 compatibility? ... Is there actually a MUCK out there still running FB 5 or are you doing this for 'good stewardship'?
Honestly backwards compatibility specifically with FB5 isn't a big concern for me -- the progs just what exposed what I figured could become a larger problem.

(In my case I got a bunch of stuff from Nameless MUCK, a server+MUF+MPI distribution forked from FB5 that probably started from or shared softcode with FB4 -- it apparently has one surviving game still using it...)

wyld-sw:
I agree that the programmer contract needs to be upheld for __version. Perhaps we could introduce some other version string/object down the line if we really need to.

I have a little bit of experience with Nameless, and it has certainly survived a long time without any update. That's very impressive.

PR #670, opened 2022-11-27
Workaround for newlines in IPv6 hostnames

Loonesbury:
"On some systems" is my vague way of saying "I was only able to test it on two systems running Ubuntu 20.04".


tanabi:
Hrm I actually want to take a shot at adjusting this.  strcpyn actually knows the length of the string that it has just copied and its return value is currently garbage; I want to change it to return the string length so that the extra strlen here isn't necessary.

The return value of strcpyn is used (for no particular reason since it just returns the char buffer you put into it) in a handful of places and that will need to be fixed, but I think in the future this will be a more useful return value for the call.  I'll have this done in a few minutes.

PR #671, opened 2022-11-28
Addresses concerns in PR #670

tanabi:
@Loonesbury


tanabi:
I'm trying to test this out now, trying to figure out how to get an IP v6 host name to go in my logs

tanabi:
Updated PR with a useful comment so I know why I did this later :)

tanabi:
Hm this didn't actually fix it.  I'm suspecting the string length is off in some direction.  Sifting through what I screwed up :)

PR #672, opened 2022-11-28
Fix off by one error

tanabi:


tanabi:
@Loonesbury this fixes it!  Had a little back and forth because I had to use the 'official' fuzzball MUCK to test it, and that feeds off master on an automatic update basis.    But it's good to go now.  Thanks for flagging the issue!

Loonesbury:
Thanks for catching that - not sure why I didn't think to check for length = 0, or for carriage returns... It's working fine now.

PR #677, opened 2022-12-18
Minor bug and compiler warning fixes

Loonesbury:
As mentioned in 6943e5af, this doesn't do much for the bulk of the compiler warning spam -- some of which I don't think is possible to reduce.

### DIFF
--- a/src/game.c
+++ b/src/game.c
@@ -822,6 +822,13 @@ process_command(int descr, dbref player, const char *command)
 
                             case 'o':
                             case 'O':
+                                /*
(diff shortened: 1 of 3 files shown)


tanabi:
All I can say about this one is good eye on that command list bug.  It's something that's obvious once you point it out, but I didn't notice it in all the times I've been through that code myself.

Loonesbury:
Thanks - I only found it because I accidentally typed `@co` after trying `@conlock` to see what it did, and then I used another command and then `@co` stopped working.

Issue #675, opened 2022-12-17
READ from MUF called on Welcome Screen causes oddities


tanabi:
This is described in detail with #674

However, the summation is, if MUF is called by MPI on the welcome screen, and that MUF has a READ call, it will cause oddities in data input and render the MUF unable to be edited without restarting the MUCK.

Forcing all MUFs called from the welcome screen to be in BACKGROUND mode would be one solution.  Fixing READ so it works appropriately is another.  Or disabling READ altogether if called from BACKGROUND.

Part of this solution may be to allower player `#-1` to run MPI / MUF, however, I think that would require a huge level of re-work.  Is it worth it?  ... Maybe.  Let's investigate and see.

tanabi:
We decided to make READ not work on the welcome screen for 7.1 and kick the can for a real solution to 7.2

PR #683, opened 2023-02-26
#675 disabled READ when MUF is called from welcome screen

tanabi:
Solves #675 at least temporarily.


wyld-sw (approved):
A good solution. The "failed check" seems to be a more foundational issue with Appveyor and conan.

PR #690, opened 2023-05-04
Bumped version to 7.1.

wyld-sw:
Added 'man propqueues' change to release notes.
Fixed minor typo.


tanabi:
@wyld-sw also need to change include/config.h VERSION define

wyld-sw:
@tanabi I think I see that included.

PR #691, opened 2023-05-06
Clarified "Program not compilable..." message 

wyld-sw:
...to "Unable to compile [[UNPARSED_OBJ]]." in the following contexts:

- viewing public words within the editor
- directly invoking a program
- calling a program from another context (propqueue / other program)

I did not test writing to the muf_errors log, as the related issue mentioned. My understanding is that do_compile takes care of writing to that log, but the "unable to compile" message happens after that.


tanabi:
LGTM.

PR #710, opened 2023-12-31
Move guest-check inside do_set, and allow general wizards to un-guest.

wyld-sw:
A little redundant in order have the guest-check near the beginning, but "set" and friends are screaming for a refactor anyway.


tanabi (approved):
Looks good to me!

PR #711, opened 2024-01-03
Unifies basic sanity checks for object creation.

wyld-sw:
This places name sanity checking closer to the point of object creation, so it can be used in all caller contexts. Player passwords are also checked in this way.

Commands affected are `@action`, `@clone`, `@create`, `@dig`, `@mcpprogram`, `@open`, `@pcreate`, `@program`, and `create` on the login screen.
Primitives affected are `COPYOBJ`, `COPYPLAYER`, `NEWEXIT`, `NEWOBJECT`, `NEWPLAYER`, `NEWPROGRAM`, and `NEWROOM`.

**Techniques**
- Passing error messages with a string pointer. Seems like an old-style idea, so let me know if there's a more preferred way. It could eventually be an error struct...
- SMALL_BUFFER_LEN = ~100~ 128. I didn't want to use another huge chunk of space for this. ~100~ 128 may be flexible enough for now - I was able to use it in several other places where buffers are also used.

**Consequences**
- `create` on the login screen can now be more detailed if the creation fails, and tp_create_fail has been removed.
- Some error messages needed to be combined. do_action and do_open refer to "an exit or action", and do_clone (as it uses create_object) refers to "create" not "clone".

**Bonuses**
- This corrects the *arrow anti-pattern* or minor *pyramid of doom* in do_open. This is far from the only place this correction can be made.
- prim_newroom now uses the nearest ABODE room as a parent (or default), like do_dig does - if the parent argument is NOTHING.

**Future Work/Discussion**
- Charging a number of pennies to create certain objects has historically been a command-only transaction, and it might make sense to keep it this way?
- It's certainly possible to check types (can only parent a room to a room, can only clone things) in this manner, though a theoretical MUF preprocessor could also check types of all primitive arguments. I guess it wouldn't hurt to do both?

### DIFF
--- a/include/db.h
+++ b/include/db.h
@@ -1900,7 +1900,8 @@ int controls_link(dbref who, dbref what);
  * With a given owner, name, and source (location).  Returns the DBREF
  * of the exit.  The 'name' memory is copied.
  *
- * Uses an error parameter to communicate the reason for failure.
+ * Uses an error parameter to communicate the reason for failure. This
+ * should be at least SMALL_BUFFER_LEN in size.
  *
  * @param player the owner dbref
  * @param name the name of the new exit
@@ -1916,7 +1917,8 @@ dbref create_action(dbref player, const char *name, dbref source, char *error);
  * With a given name and owner/creator.  Returns the DBREF of the program. 
  * The 'name' memory is copied.  Initializes the "special" fields.
  *
- * Uses an error parameter to communicate the reason for failure.
+ * Uses an error parameter to communicate the reason for failure. This
+ * should be at least SMALL_BUFFER_LEN in size.
  *
  * @param player the owner dbref
  * @param name the name of the new program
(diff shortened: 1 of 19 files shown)


tanabi:
This is a big 'un!  I will dive in full tomorrow :)

include/db.h
```diff
@@ -1900,24 +1900,30 @@ int controls_link(dbref who, dbref what);
  * With a given owner, name, and source (location).  Returns the DBREF
  * of the exit.  The 'name' memory is copied.
  *
+ * Uses an error parameter to communicate the reason for failure.
```
tanabi, line 1903: I would rephrase this slightly:

Uses an error parameter to communicate the reason for failure.  This should be at least SMALL_BUFFER_LEN in size.  Obviously this applies to all locations with a similar comment.

tanabi (changes requested):
Looks mostly good.  I think a note about the assumed size of error will be useful; it is unlikely an undersized buffer would be used, but it's best to not be ambiguous.

include/db.h
wyld-sw: Good idea! Commit has been pushed.

PR #713, opened 2024-01-08
Unifies setting and clearing lock properties.

wyld-sw:
This attempts to combine the locking implementations used in the codebase.

**Reunification**

Most of the lock commands (e.g., `@lock`, `@ownlock`) use set_standard_lock for their implementation. `@unlock` and `SETPROPSTR` did not; this is now corrected.

**CLEARLOCK, GETLOCK, and SETLOCK**

These macros were used for the main object lock only; they are now extended to act on any type of lock.

CLEARLOCK now removes the property (instead of setting it to TRUE_BOOLEXP) as other mechanisms do. It also continues to update an object's modification time.

GETLOCK can now be used in all cases that get_property_lock was called directly. Is it redundant? Probably - but then so are  lot of other GET* macros.

SETLOCK is now used in the refactored \_set\_lock (discussed below). It also continues to update an object's modification time.

**\_set\_lock**

This helper function is now called by set_standard_lock and prim_setlockstr, with a "silent" argument to determine if messages are displayed to the user. Lock parsing errors continue to be displayed no matter what. I was hoping to use a smaller method signature, but them's the breaks.

**Thoughts**

I wonder if we should introduce `SETCHLOCK`, `SETCONLOCK`, `SETFLOCK`, `SETLINKLOCK`, `SETOWNLOCK`, and `SETREADLOCK` - since the standard messages have their own "set" aliases.

_Suggestions welcome._

### DIFF
--- a/include/db.h
+++ b/include/db.h
@@ -303,25 +303,27 @@ extern char match_cmdname[BUFFER_LEN];
  * A thin wrappera round @see get_property_lock
  *
  * @param x the object to look up
+ * @param y the lock to look up
  * @return the struct boolexp object or TRUE_BOOLEXP
  */
-#define GETLOCK(x)      (get_property_lock(x, MESGPROP_LOCK))
+#define GETLOCK(x,y)      (get_property_lock(x, y))
 
 /**
  * Sets a lock from a boolexp structure
  *
  * Typically you would take the return value from @see parse_boolexp
- * and then use it as the 'y' parameter here.  The object modification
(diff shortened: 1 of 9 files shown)


tanabi (approved):

tanabi:
I really like the consolidation here, looks quite good.  As for the idea for other macros ... I think that's a pretty good idea really.  Though not necessarily pressing to do right now, we could backlog a ticket for it.  I'm going to approve this ohe and merge for now.

PR #714, opened 2024-01-09
Error prim fixes, GAUSSIAN simplification, and style improvements

wyld-sw:
**Non-cached `GAUSSIAN`**

The current `GAUSSIAN` caches one of the generated random numbers for a potential next call, as does the traditional algorithm. A comment suggested we may not need to do this, and I would tentatively agree based on my limited testing. Caching does certainly have an impact, but I believe a program would exhaust the instruction limit before that impact is noticeable (even with a much larger instruction limit).

My limited testing included using the undocumented `{timing}` function to loop through several thousand calls (`100.0 5000.0 GAUSSIAN`), with both implementations (with and without caching). Without caching, I hit the instruction limit in less than 0.1 seconds. Now that was just my single-player test server, so I may be jumping to conclusions.

This is mostly the reason to review this PR, but please point out any problems with the other parts too!

**Errors within errors**

`CLEAR_ERROR` , `ERROR_STR`, `IS_SET?`, and `SET_ERROR` now function as expected (again) when used with a string argument. I made a commit years ago that inadvertently broke this.

Some of these primitives have been rewritten to use less code.

**Style, scope, and sanity**

One-line if-blocks now use braces.

Many variables have been moved closer to their access points, including globals. Except for oper1-oper4 because they're special (for now).

Redundant and ill-placed sanity checks that were missed in a recent PR for p_misc.c have been combined.

### DIFF
--- a/src/p_error.c
+++ b/src/p_error.c
@@ -10,34 +10,13 @@
 #include "inst.h"
 #include "interp.h"
 
-/*
- * TODO: These globals really probably shouldn't be globals.  I can only guess
- *       this is either some kind of primitive code re-use because all the
- *       functions use them, or it's some kind of an optimization to avoid
- *       local variables.  But it kills the thread safety and (IMO) makes the
- *       code harder to read/follow.
- */
-
 /**
  * @private
  * @var used to store the parameters passed into the primitives
  *      This seems to be used for conveinance, but makes this probably
- *      not threadsafe.
+ *      not threadsafe. Currently used in the abort_interp macro.
  */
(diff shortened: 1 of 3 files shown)


tanabi (approved):
This is some really good work, I like the changes.  And removing those statics I think is a good thing; not only will it not appreciably impact the server load on a modern machine, but it's such an edge case weird call.  I have a hard time believing there's too many MUFs out there that use it.

I added an issue #715 to address teh abort_interp problems.  It occured to me that C++ is basically designed to solve these issues, so it may be a good use case for a C++ re-write of the MUF system.

PR #718, opened 2024-01-12
Add {flag?} MPI function

wyld-sw:
This adds an MPI counterpart to the `FLAG?` MUF primitive. The check logic has been extracted into its own function.

**Functionality**

Both MPI and MUF versions check a single flag, which can be a prefix of a flag alias (regardless of object type). The check can also include any number of "!" characters on the left-hand side of the string for alternating negation.

This may be the first object-facing MPI function I've contributed, so let me know if I missed something there.

**Consistency**

This check now recognizes BOUND, HIDE, SETUID, and XPRESS as flag aliases. `SET` also was changed to recognize them.

**Minor change**

The code for `FLAG?` currently:
- allows "" as the flag to check, and eventually returns 0.
- allows "U" as the flag to check, and eventually returns 0.
- does not allow "!" as the flag to check, and aborts with "Unknown flag"

This seems inconsistent, as the first two are similar conditions to the third. I've changed the third case to return 0 (with a bonus of returning 0 more quickly in the first case). I can see a corner case where one would want to abort when checking for "!", but it doesn't seem that compelling. Feel free to disagree, and I can change it to pass the error condition back to `{flag?}` and `FLAG?`.

### DIFF
--- a/src/mfuns2.c
+++ b/src/mfuns2.c
@@ -3460,3 +3460,37 @@ mfn_width(MFUNARGS)
 
     return buf;
 }
+
+/**
+ * MPI function that returns if an object if it has the given flag (re)set.
+ * 
+ * @see has_flag
+ * 
+ * @param descr the descriptor of the caller
+ * @param player the ref of the calling player
+ * @param what the dbref of the trigger
+ * @param perms the dbref for permission consideration
+ * @param argc the number of arguments
+ * @param argv the array of strings for arguments
+ * @param buf the working buffer
+ * @param buflen the size of the buffer
+ * @param mesgtyp the type of the message   
+ * @return string parsed results
+ */
+
+const char *
+mfn_flagp(MFUNARGS)
+{
+    dbref obj = mesg_dbref_local(descr, player, what, perms, argv[0], mesgtyp);
+
+    if (obj == PERMDENIED || obj == AMBIGUOUS || obj == UNKNOWN || obj == NOTHING ||
+        obj == HOME) {
+        return "0";
+    }
+
+    if (has_flag(obj, argv[1])) {
+        return "1";
+    } else {
+        return "0";
+    }
+}
--- a/src/mfuns2.c
+++ b/src/mfuns2.c
@@ -3485,7 +3485,7 @@ mfn_flagp(MFUNARGS)
 
     if (obj == PERMDENIED || obj == AMBIGUOUS || obj == UNKNOWN || obj == NOTHING ||
         obj == HOME) {
-        return "0";
+        ABORT_MPI("FLAG?", "Failed match. (arg1)");
     }
 
     if (has_flag(obj, argv[1])) {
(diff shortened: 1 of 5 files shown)


src/mfuns2.c
```diff
@@ -3460,3 +3460,37 @@ mfn_width(MFUNARGS)
...
+ * @param mesgtyp the type of the message   
+ * @return string parsed results
+ */
+
+const char *
+mfn_flagp(MFUNARGS)
+{
+    dbref obj = mesg_dbref_local(descr, player, what, perms, argv[0], mesgtyp);
+
+    if (obj == PERMDENIED || obj == AMBIGUOUS || obj == UNKNOWN || obj == NOTHING ||
```
tanabi, line 3486: It looks like a "failed match" error is the more common way to handle this.  Here's a sample I found in another primitive:

```
if (obj == AMBIGUOUS || obj == UNKNOWN || obj == NOTHING || obj == HOME) {
        ABORT_MPI("FLAG?", "Failed match. (arg1)");
}
```

src/p_db.c
```diff
@@ -1071,7 +1071,8 @@ prim_set(PRIM_PROTOTYPE)
                 || string_prefix("autostart", flag)
                 || string_prefix("abate", flag))
             tmp = ABODE;
-        else if (string_prefix("builder", flag))
+        else if (string_prefix("builder", flag)
```
tanabi, line 1074: Perhaps we should note in both this code and in has_flag that any change needs to be made in both places?

Ideally, we'd unify these 2 case statements in some way since it looks like this has a lot of overlap with has_flag's code.  However, there's also such a thing as optimizing yourself until you're weary, so if you rather put a TODO and a note that both places need to be changed, I wouldn't judge you :)

If you wanted to go the extra mile, though, I'd say make a str_to_flag call that has this case statement, then both has_flag and prim_set can use it (along with anywhere else this mess lives).

tanabi (changes requested):
One necessary change, one suggestion/comment.  Looks really good! :)

src/mfuns2.c
tanabi: OH ... and obj==PERMDENIED in that list, too. Whoops :)  Basically, keep your if statement, but change the return to ABORT_MPI

src/p_db.c
wyld-sw: str_to_flag is definitely coming - just wanted to see if I could be clever with ALL the flag checking. I can certainly add str_to_flag to this PR in the meantime so it definitely doesn't wait for a unified process.

wyld-sw: I think both changes are made. Please review; I think everything's in order...

I'm not too thrilled about the duplicate "truewizard" check, but it seemed like the best way (right now)? to preserver the different meanings for `SET` and `FLAG?`.

do_set is on notice as needing a change, too, but it conflates the str_to_flag logic with the restricted function (while introducing its own restrictions) and will take more thought.

tanabi (approved):

Issue #46, opened 2016-12-24
User passwords stored using insecure algorithm


ghost:
All user passwords are currently stored according to the following algorithm:

```
base64(md5(password))
```

Using MD5 is, these days, little better than obfuscation, rather than proper security. Combined with a lack of salting, it is trivial to recover user passwords with access to the database.

Salting passwords prior to storage would greatly increase the time to recovery for dictionary and brute-force offline attacks, though with MD5 as the underlying hash, it would still be well within the capability of the average attacker.

Ideally, MD5 would be replaced with a proper key derivation function (bcrypt, PBKDF2, et al.), or at least a multi-round SHA2 digest.

Both of these (salting, algorithm) could be implemented as compile-time options so as not to automatically modify a database in such a way that compatibility is broken with prior versions.

kyanha:
I agree, this is a problem.

I think that bcrypt or whatever obfuscation system should be compiled in by default (so as to always be able to use non-backward-compatible databases), with a God-level tunable parameter to make password storage be non-backward-compatible when '@password' and '@newpassword' are used.

wyld-sw:
Agreed.

sypherspace:
@kyanha is probably referring to `@password` and `@newpassword`. Gotta love Markdown.

digitalcircuit:
On FreeBSD/Linux/etc, there's the [```crypt()``` function](https://www.freebsd.org/cgi/man.cgi?query=crypt&sektion=3 ) which offers a semi-standard way of storing passwords, e.g. the system uses it for user logins and [PHP uses it for password hashing](https://secure.php.net/manual/en/function.password-hash.php ).  This would simplify upgrading to more secure password hashes in the future as the version is stored as part of the hash string, and it helps avoid mistakes in implementation.

I'm not sure about how to handle this on Windows, unfortunately.  Based on some discussion with ```Freenode/AllanJude```, the FreeBSD implementation should be portable to Windows, but I haven't tried that and don't know what complications may arise.

*Aside, any solution here would probably help Quassel, too, where I first encountered the topic - [comment by ```@allanjude``` on ```quassel/quassel``` pull request 110](https://github.com/quassel/quassel/pull/110#issuecomment-227048493 ).*

sypherspace:
The `crypt()` function would probably be the easiest route forward. It would be easy to parse old vs. new passwords in the database: Does the password string start with a `$`? If yes, it's `crypt()`, If no, it's old-style MD5.

[...]

PR #720, opened 2024-01-28
#46 Added secure password algorythm

tanabi:
This adds a strong hash to the password algorythm re. #46 

Note that, for compatability purposes, it does not automatically rehash the database.  This gives the admin the chance to `@tune tp_legacy_password_hash=true` if they don't want this update for some reason.  Admins that wish to roll this out should advise their players to use `@password` to rehash their passwords.


wyld-sw:
Good stuff. Thanks!

PR #729, opened 2024-04-12
Fix for buffer overflow issue #728

### LINKED ISSUE #728: New password hashing algorhithm causes crash on login
Finally created my own Github account just report this, instead of nudging poor @foxbird to note these things for me for years, hee.

The new password hashing algorithm (#720) causes an immediate hard crash on login.  Very easy to reproduce:

Start MUCK.  Log in.  Change password (with legacy_password_hash at default, so the new password is hashed with the new algorithm).  QUIT.  Reconnect.  Immediate hard crash on entering the `connect <char> <password>` command.  Even tested it using the minimal and starter databases, to make sure it was nothing to do with my existing database.

Restart, `@tune legacy_password_hash=yes`.  Repeat all other steps.  No crash, everything works normally.

**Of note:** I reviewed the logs/status file.  When using the new algorithm and attempting to log in after changing a character's password (i.e., after the password has been hashed to the new algorithm), the logs/status file shows

> CONNECTED: Room Zero(0), descriptor 9, from localhost

instead of the correct character and dbref.  This entry occurs in the instant before the fbmuck process crashes.  When using the OLD algorithm, connecting a character looks normal 

> CONNECTED: Keeper(2), descriptor 9, from localhost

So it would appear to be something in the login code isn't handling the new hash, and screws up who/what is actually being logged in, attempting to literally log in as Room Zero.  Assuming the logs/status file is correct, anyway.

**Secondarily**: also when using the new algorithm, the `@newpassword` command likewise misreports the user in the logs/status file.  However, unlike the CONNECTED line, the newpassword line gets the actual dbref correct.  For example, using the starter database, connecting as One, and using `@newpassword` to change Keeper's password, logs/status shows

> NEWPASS'ED: Room Zero(2) by One(1)

When using the OLD algorithm (`@tune legacy_password_hash=yes`), it correctly shows

> NEWPASS'ED: Keeper(2) by One(1)

Since `@password` doesn't add an entry to logs/status the way `@newpassword` does, I have no idea if it's having the same object reference weirdness or not.  Nonetheless, there's all kinds of issues with the processing of the new algorithm.  These things are just what I've found so far, more or less immediately upon starting up with a new git pull.

Sure hope it's something weird with my setup (bog standard Debian Bullseye), and not that something as basic as _logging in after changing the password_ wasn't tested before this new code was committed.  c.c

tanabi:
This resolves #728 

This works without an issue with my compiler, which is why I didn't catch it.  Since some folks may be already running this version with the password problem, and thus have 128 character instead of 127 character passwords, I made sure the password check mechanism will check based on the length of the hash.

### DIFF
--- a/src/fbmath.c
+++ b/src/fbmath.c
@@ -584,26 +584,6 @@ rnd(void *buffer)
  *
  *********************************************************************/
 
-#ifdef USE_SSL
-static void
-PBKDF2_HMAC_SHA_512(const char* pass, const unsigned char* salt,
-                       int32_t iterations, uint32_t outputBytes,
-                       char* hexResult)
-{
-    unsigned int i;
-    unsigned char* digest;
-
-    digest = (unsigned char*)malloc(outputBytes);
-
-    PKCS5_PBKDF2_HMAC(pass, strlen(pass), salt, strlen(salt), iterations,
-                      EVP_sha512(), outputBytes, digest);
-    for (i = 0; i < outputBytes; i++)
-        sprintf(hexResult + (i * 2), "%02x", 255 & digest[i]);
-
-    free(digest);
-}
-#endif
-
 /**
  * Generate a PBKDF2 password hash with the given password and salt.
  *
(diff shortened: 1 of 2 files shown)


davidatpurdue:
Great catch and teamwork!

wyld-sw:
Great catch and teamwork!

PR #742, opened 2025-02-05
Fix for #741

### LINKED ISSUE #741: Empty Password allows logging in as user
I compiled and ran the server both with the docker-compose.yml as well as manually with make. As long as I type an incorrect password when connecting the server will detect it as invalid, but if no password at all is provided it allows logging in as the user (apparently unconditionally).
Is this meant to be the case, or is it in some kind of debug mode by default?

casuallyblue:
This should at least partially fix #741
It makes sure that the hash of the entered password isn't empty. My assumption is that if the hash isn't zero it should always be a specific length hence a single check for zero length should be enough to fix the issue, if this isn't the case then another check might be necessary.

### DIFF
--- a/src/player.c
+++ b/src/player.c
@@ -81,6 +81,7 @@ check_password(dbref player, const char *password)
 {
     char md5buf[128];
     int len_hash = 0;
+    int len_processed = 0;
     int len_salt = 0;
 
     const char* salt;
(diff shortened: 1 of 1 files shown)


tanabi (approved):

tanabi:
This is exactly what I was going to do.

PR #747, opened 2025-05-08
Boolexps: Recognize MPI and limit recursion

### LINKED ISSUE #86: lock parsing/unparsing/etc. can use a lot of stack space
The lock parsing/evaluation/unparsing code is recursive and the only limits on the number of levels are based on what can be fit in a (~8KB) lock string. It might be wise to have some explicit depth limit rather than have a way that users might be able to make the MUCK run out of stack space.

### LINKED ISSUE #320: MPI Support for Boolexp (Locks)
Okay, as a player, I've wanted this for a really long time.  Even as a server admin, this would be super handy.  I would love for @ lock to parse MPI.  There's no particular reason for it not to, and in fact, this is a fairly easy change.  If anyone wants to contribute to Fuzzball but has been shy about it, this is an easy(ish) one to implement.

1. Create a BOOLEXP_MPI define in boolexp.h
2. Add a char* to struct boolexp for the MPI text to parse.  Bonus points if you turn thing/prop_check/mpi into a struct union -- that's more work since you'll have to convert the code in boolexp.c to use a union, but it actually reduces the memory footprint of struct boolexp instead of increasing it :)  If you chose not to go for the bonus points, please put an @ TODO comment reminding us to come back and turn it into a union later.
3. Add support for the MPI string to copy_bool
4. In eval_boolexp_rec add a case BOOLEXP_MPI and have it evaluate the MPI.  Make sure to run it in PREEMPT.  There's plenty of example codes of how run MPI.  The results of the evaluation should be either "0" or anything else -- anything else will be counted as true, similar to how the MUF evaluation under BOOLEXP_CONST works.
5. Updat eparse_boolexp_F to look for a '{' in the big case statement there.  The way this should work is roughly as follows:
```
case '{':
  int brace_count = 1;
  char* mpi_start = *parsebuf;
  (*parsebuff)++;
  for( ; *parsebuf ! = '\0' && brace_count ; (*parsebuf) ++) {
      if *parsebuf == '{' then brace_count++;
      if *parsebuff == '}' then brace_count --;
   }
   if *parsebuf == '\0' then return TRUE_BOOLEXP <-- did not find last brace
   else create boolexp node for MPI with string that starts at mpi_start and ends at *parsebuf
   (*parsebuf)++   // not sure if this is needed or not.
   return new boolexp node
```
6. Update size_boolexp to support BOOLEXP_MPI
7. Update free_boolexp to support BOOLEXP_MPI
8. Update unparse_boolexp1 to support BOOLEXP_MPI
9. Update the @ lock documentation ... muckhelp.raw

That should be it ... most of those are pretty small changes.  If nobody beats me to it, I'll do it when I'm done with the documentation project.

wyld-sw:
Adds BOOLEXP_MPI as a node type for locks, and re-implemented values as a union type (resolves #320).

Adds recursion limit for parsing locks (resolves #86) - currently 50, please change if needed.

This seems to work successfully during local testing, but please review.

### DIFF
--- a/include/boolexp.h
+++ b/include/boolexp.h
@@ -14,12 +14,15 @@
 #include "config.h"
 #include "props.h"
 
+#define MAX_LOCK_DEPTH 50  /**< Recursion limit */
+
 /* These are the different elements of a boolean expression */
 #define BOOLEXP_AND 0    /**< AND is the default */
 #define BOOLEXP_OR 1     /**< OR expression */
 #define BOOLEXP_NOT 2    /**< NOT expression */
 #define BOOLEXP_CONST 3  /**< CONST expression */
 #define BOOLEXP_PROP 4   /**< property expression */
+#define BOOLEXP_MPI 5    /**< MPI expression */
 
 /* True is a null boolean expression */
 #define TRUE_BOOLEXP ((struct boolexp *) 0)   /**< Boolean TRUE expression */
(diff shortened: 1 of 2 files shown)


tanabi (approved):
This is some really good work, I like how you implemeneted this.  Well done!

PR #758, opened 2026-04-26
Attempts a refactored flag subsystem.

wyld-sw:
# Draft PR: Flag Logic Refactor & Multi-Flag Support

This is a significant refactor. I appreciate any reviewer taking the time to wade through the changes.

## Summary
This PR refactors and relocates flag checking and presentation logic into a centralized subsystem. The primary goal is to make the system manageable and support a flag (re)setting refactor, while introducing powerful new multi-flag checking capabilities.

### Key Additions
* **Multi-Flag Checking:** Players can now check multiple flags at once (e.g., `"L!A"`) via:
    * The `FLAGS?` MUF primitive.
    * The `{flags?}` MPI function.
* **Flag Subsystem:** Centralized logic in `include/flags.h` and `src/flags.c` using lookup tables for bit values, flag symbols, and type aliases.

---

### Renames for Clarity
| Old Name | New Name |
| :--- | :--- |
| `Typeof` | `OBJECT_TYPE` |
| `MLevRaw` | `OBJECT_MLEVEL` |
| `MLevel` | `OBJECT_EFFECTIVE_MLEVEL` |
| `Plevel` | `OBJECT_PRIORITY` |

### New Public Functions & Defines
* `FLAG_CHECK`: Returns true if a single-character flag is set.
* `FLAG_VALUE`: Returns the bit value of a single-character flag.
* `flag_init`: Initializes lookup tables from an array of structs.
* `flag_eval`: Underpins `FLAG?` and `{flag?}`; recognizes symbols, aliases, and "truewizard".
* `flag_eval_pattern`: New multi-flag checker; recognizes flag symbols, type symbols, and MUCKER levels.
* `flag_list` / `flag_list_verbose`: Generates single-symbol or "wordy" flag lists.
* `flag_unparse_object`: Revised `unparse_object` utilizing the new lookup tables.
* `object_type_name`: Returns the uppercase name of an object's type.

---

## Discussion Points
These changes might be for the better, but they introduce specific logic shifts. Please review the following:

### 1. Type-Strict Lookups
The lookup tables now use the object's type as an index. Checking if a Program is `SILENT` or an Exit is `AUTOSTART` will now return **false**. 
* **Impact:** This may break legacy programs relying on "type confusion."
* **Decision:** Should we maintain the loose legacy logic (at a performance cost) or enforce strictness?

### 2. Symbols & Aliases
* **The "T" Symbol:** I've introduced `T` as the type symbol for "Things." Note that this overlaps with the "truewizard" alias in single-flag checks, but those don't recognize type symbols.
* **MUCKER Levels:** Levels are checked via `"PW3"`. "M2" and "M3" work as aliases, but only since they both include the "M" bit. Currently, `"M1"` fails because it is technically `N`. 
* **The M0 Case:** Checking for Mucker level 0 might not ever return true.
* **Proposal:** Should we add code to make `M1` intuitive, or perhaps remove `M` from the short flag list entirely to manage search expectations?

### 3. Internal Flags (Untouched)
I have excluded the following internal flags from this refactor for now. I am considering moving some of these out of the flag system.

| Flag | Purpose | Current Status |
| :--- | :--- | :--- |
| `INTERACTIVE` | Player is in MUF editor | **Searchable** via flag checkers. |
| `INTERNAL` | Player sees line numbers / Program is being edited | Saved to disk; needs prop-based migration. |
|`LISTENER`| Determines if object has a listener propqueue | Re-applied on database load. |
| `OBJECT_CHANGED` | Delta-dumping | Historical utility. |
| `READMODE` | Player in a READ state | Runtime state only. |
| `SANEBIT` | Database sanity tools | Cleanup/Sweep utility. |

---

## Miscellaneous Fixes
* **Style:** Removed trailing spaces in `.c` and `.h` files; fixed control flow spacing.
* **Refactors:** Simplified `do_examine` unparsing and `do_dig` flag inheritance.
* **Safety:** Reduced compiler warnings by patching a few potential buffer overflows.

---

## Testing Status
Testing has been successful in my local environment, but given the scope of the refactor, I welcome any edge cases I may have missed.

### DIFF
--- a/docs/man.txt
+++ b/docs/man.txt
@@ -126,12 +126,12 @@ E's
 
 F's
   fabs               fail               fg_mode            findnext
-  firstdescr         flag?              flags              float
-  float?             floor              fmod               fmtstring
-  fmttime            for                force              force_level
-  forcedby           forcedby_array     foreach            foreground
-  fork               frand              ftostr             ftostrc
-  fulldepth          
+  firstdescr         flag?              flags              flags?
+  float              float?             floor              fmod
+  fmtstring          fmttime            for                force
+  force_level        forcedby           forcedby_array     foreach
+  foreground         fork               frand              ftostr
+  ftostrc            fulldepth          
 
 G's
   gaussian           getlink            getlinks           getlinks_array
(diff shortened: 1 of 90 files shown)


tanabi (approved):

tanabi:
Hoooo boy, this was a lot to review, but it looks quite solid to me. a lot of it is just changing the the calling pattern to the new way. I do like how it's implemented. Do we wanna wait and see if anyone else wants to review it or just smash the merge button?

wyld-sw:
I'm fine with letting it sit for a bit in case another reviewer comes along. Feel free to merge when you think is right.

wyld-sw:
I just pushed a fix to @find/@owned/etc. 
I also added the function object_type_name_mpi and used it in `{type:}` and `{istype:,}`

wyld-sw:
Refactored a bit more, and `@stats` now uses the new type system. Directly accesses the struct, but I think that's fine for now.

Oddly, it looks like mesg_dbref_raw() never returns NOTHING, AMBIGUOUS, or HOME (preferring UNKNOWN). I think I verified that with FM, too, so it hasn't been a logic change I've seen. What this seems to mean is that all that checking for these values is unnecessary?

If we do allow this, we should add a check for NIL as well. If we leave it alone (my preference, I think), we might be able to remove several checks.

wyld-sw:
I removed M from the short flag lists, to match how flag pattern searches work. Hopefully it'll be less jarring than seeing T for things (something at least one MUCK already does!)

I also added documentation for the new MPI and MUF functionality.

Merging this, but feel free to mention any problems you see, and I'll circle back.

PointsDragon:
8.651 flags.c: In function 'flag_unparse_object':
8.651 flags.c:351:21: error: implicit declaration of function 'can_see_flags'
Fails to compile with warnings enabled - I thought we ALWAYS have all warnings enabled? :) [-Wimplicit-function-declaration]

Issue #743, opened 2025-02-05
Setting a password when compiled with SSL support results in inability to log in


casuallyblue:
setting a new password for a user, either through `@password oldpass = newpass` or via `@pcreate player = pass` in the default configuration of the docker container results in only `$1$M$` being written to the database no matter what the password chosen is.

Currently looking into exactly why, but I haven't discovered the source of the issue as of yet.

wyld-sw:
Addressed by #761

tanabi:
@casuallyblue Dockerhub is updated with a 7.2.3 that fixes this.

PR #761, opened 2026-06-07
Fix Password Weirdness

### LINKED ISSUE #744: Possible instability related to @passwords.
Okay, this is the third time I have witnessed this, and now I have run into staff and now users who have run into this.

### Scenarios:
**Setting one:**
Login screen create new user, they create/pass no problem and log right in.
They log out, then log back in again with the same user/pass they still have in their muck buffer.. and the password does not work.

Admin manually sets their password to the same password they just tried, and it works,. user can now log in normally.

This does not happen every time, it seems to happen at random.

**Setting two:**
Existing user has lost their password. Staff verify it is them, and change their password with '@ newpassword'
Password reports it is changed successfully.
User can not log in with the password.
Admin uses their input history to run the exact same command over again..
Now the password works.

**Setting three:**
User decides to change their password themselves. 
They change it, it reports password changed. 
They go to log back in, password does not work.
old password also does not work.
User cannot log in, admin is called in to change it. 
User can now log in.

In all these cases, the user input was confirmed/logged, and in all these cases, the password setting just.. failed, somehow, internally, without any error anywhere we can see.

You can test this with two characters in our case. you can log in with a wizbit, and a normal character, and keep setting their password to the same thing over and over again, and have the second login reconnect each test.

It seems to just randomly.. fail, and set some kind of garbage unknown password.

I actually ran into this months ago, but I just knee-jerked and assumed I had flubbed the input somewhere. But now I keep running into it over and over again, and finally just.. tested it. 

There is a problem buried here somewhere.

tanabi:
This resolves #744 and #743

It turns out that we could randomly wind up with 'accidentally' too short strings based on a random factor which is why we sometimes see this and sometimes don't.  The pool of characters that we were generating salts from could occasionally include null and cause truncation.

I've tested this a bit locally. It's difficult to know for 100% sure this is fixed because it's a random factor, but this seems the most likely problem. I haven't had this happen since doing the change. I encourage those that experience this problem to update and see if it resolves things. I'll be updating my own MUCK to this version and updating the docker stuff once we've got this merged.

### DIFF
--- a/docs/man.txt
+++ b/docs/man.txt
@@ -126,12 +126,12 @@ E's
 
 F's
   fabs               fail               fg_mode            findnext
-  firstdescr         flag?              flags              float
-  float?             floor              fmod               fmtstring
-  fmttime            for                force              force_level
-  forcedby           forcedby_array     foreach            foreground
-  fork               frand              ftostr             ftostrc
-  fulldepth          
+  firstdescr         flag?              flags              flags?
+  float              float?             floor              fmod
+  fmtstring          fmttime            for                force
+  force_level        forcedby           forcedby_array     foreach
+  foreground         fork               frand              ftostr
+  ftostrc            fulldepth          
 
 G's
   gaussian           getlink            getlinks           getlinks_array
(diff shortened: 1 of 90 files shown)


wyld-sw:
Makes sense, and seems to work for me. Thanks!

PR #769, opened 2026-06-17
#763 Fix pointer math problem that was causing core

### LINKED ISSUE #763: Server panic (Signal 6 / SIGABRT) during examine with property paths on modern GCC optimization levels
When running the server compiled with standard optimization flags (-O2 or -O3) on a modern system (GCC 11+ / Ubuntu), using the examine command with a property path (e.g., ex me=/ or ex me=/_/de) causes an immediate server panic: PANIC: BAILOUT: caught signal 6.

The crash occurs deep inside listprops_wildcard() due to how modern aggressive compiler optimizations handle the legacy pointer arithmetic/loops used for property directory traversal.

Environment:

OS: Ubuntu (Linux)

Compiler: GCC 13.3.0

Workaround / Fix:
Compiling the server with optimizations completely disabled (-O0 or CFLAGS="-g -O0") completely resolves the issue, and the examine command functions normally without crashing.

tanabi:
This fixes #763 

This was incorrectly attributed to a compiler optimization; the optimization causes the problem to surface, but the real problem is that we're passing 'bufsize' which is the size of 'buf' as the size to 'mybuf' which is actually smaller than 'mybuf' so that causes a segfault. But for whatever reason, only when optimizing.

Oops.

### DIFF
--- a/src/property.c
+++ b/src/property.c
@@ -1312,7 +1312,7 @@ displayprop(dbref player, dbref obj, const char *name, char *buf, size_t bufsiz)
         blesschar = 'B';
 
     pdflag = (PropDir(p) != NULL);
-    snprintf(mybuf, bufsiz, "%.*s%c", (BUFFER_LEN / 4), name,
+    snprintf(mybuf, sizeof(mybuf), "%.*s%c", (BUFFER_LEN / 4), name,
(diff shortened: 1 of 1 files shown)


PointsDragon (commented):
Approved.

PR #772, opened 2026-06-24
Fix for #771 to limit propdir recursion to 100 subdirs

### LINKED ISSUE #771: Immediate Server Crash Upon Iterating Through Ultra-Deep Propdir Levels
So some wonderfully lovely person has taken a liking to signing on to my server whatever way they can (guess I'll need to close character creation for a bit) and deliberately triggering a crasher bug.

While it causes a hard (no save dump) crash, at least the log captures what they do to trigger it.  They set a prop on themselves that's ultra-deep in subprops, and then "ex me=/**" to iterate all their props.  Causes an immediate crash.

I thought this might be related to the recently-closed #763 but I have updated my server version to the latest git pull which had closed that issue, and this crasher bug still exists.

I hesitate to paste the log entry directly into this issue because it'll give a precise example of exactly how to crash presumably most any FB7 instance (give or take, maybe even older, no idea where and when this was introduced).  Though my description is likely enough for anyone to reproduce and abuse it anyway.  It's trivially easy to trigger, so this is kinda.. not good!

tanabi:
This fixes #771 and adds a recursion limit to make sure `ex me=/**` or similar can't crash the MUCK. Further, it prevents SETTING a prop that deep to begin with. The default limit is 100 deep which is plenty deep for all reasonable purposes.

### DIFF
--- a/src/look.c
+++ b/src/look.c
@@ -456,10 +456,12 @@ do_look_at(int descr, dbref player, const char *name, const char *detail)
  * @param thing the thing we are checking props on
  * @param dir root dir to search from - used for recursion -- pass "" for root
  * @param wild the pattern to search for.
+ * @param depth current propdir nesting depth - pass 0 for the root call
  * @return int number of props returned.
  */
 static int
-listprops_wildcard(dbref player, dbref thing, const char *dir, const char *wild)
+listprops_wildcard(dbref player, dbref thing, const char *dir, const char *wild,
+                   int depth)
 {
     char propname[BUFFER_LEN];
     char wld[BUFFER_LEN];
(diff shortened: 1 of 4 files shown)


src/look.c
```diff
@@ -470,6 +470,25 @@ listprops_wildcard(dbref player, dbref thing, const char *dir, const char *wild)
...
+     */
+    {
+        int depth = 0;
+
+        for (const char *s = dir; *s; s++) {
+            if (*s == PROPDIR_DELIMITER)
+                depth++;
+        }
+
+        if (depth >= MAX_PROPTREE_DEPTH)
```
PointsDragon: So, to leave partial functionality for 'older' databases or for a database that someone @tunes too shallow, rather than do an O(n) forward count, we can target where the issue actually lies - add a default param to this function which is our call-depth, and then use that at line 526 here to do our cut-off.  We can have the player output for a truncated depth then have its own message as well.

PointsDragon (approved):
LGTM

tanabi:
Points gave a thorough review which we discussed, so I'll go ahead and merge this one as we have a "customer" waiting for it.

PR #777, opened 2026-07-01
Set fix - Guest and player vs. program flag actions.

PointsDragon:
Please review this carefully.  This fixes the issue where guests were allowed to set/unset props and reverted back to not being able to use the set command.  Additionally, unable_to_set_flag had its prototype changed to help differentiate between player vs. program action and the logic internally was updated to better conform to original/expected behavior.  I hope.  I also removed what looked to be a double definition of the function.

### DIFF
--- a/src/flags.c
+++ b/src/flags.c
@@ -489,7 +489,8 @@ str_to_flag(const char *flag_string)
  * @param error[out] error message, if any
  * @return boolean 1 if restricted from setting flag, 0 if okay to set.
  */
-int unable_to_set_flag(dbref player, dbref puid, int mlev, dbref thing, object_flag_type flag, bool value, char *error);
+int
+unable_to_set_flag(dbref player, dbref puid, int mlev, dbref thing, object_flag_type flag, bool value, char *error)
 {
     /* Wizard check logic. */
     bool wizlev = (player != NOTHING && Wizard(OWNER(player))) || (puid != NOTHING && mlev == 4);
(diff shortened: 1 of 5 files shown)


wyld-sw:
Looks good from here. @tanabi, please review when you have time in case I missed anything and feel free to merge. Thanks!

src/flags.c
```diff
@@ -467,30 +467,35 @@ str_to_flag(const char *flag_string)
...
  * @param mlev the effective mucker level we are checking permissions for
  * @param thing the thing we want to interact with
  * @param flag the flag we wish to interact with
  * @param value whether the desired state is on or off
  * @param error[out] error message, if any
  * @return boolean 1 if restricted from setting flag, 0 if okay to set.
  */
-int
-unable_to_set_flag(dbref player, int mlev, dbref thing, object_flag_type flag, bool value, char *error)
+int unable_to_set_flag(dbref player, dbref puid, int mlev, dbref thing, object_flag_type flag, bool value, char *error);
```
tanabi:
```suggestion
int
unable_to_set_flag(dbref player, dbref puid, int mlev, dbref thing, object_flag_type flag, bool value, char *error)
```

I don't love this coding convention but it's used literally everywhere :D
Also there was a trailing ; on the end here which ... probably won't compile ? Hmm.

tanabi (changes requested):
Style fix and possible compile error or at least oddity :)

PR #778, opened 2026-07-03
Minor TODO cleanup

### LINKED ISSUE #779: Commit 9462db5 Causes Compile Failure
Hey @PointsDragon I'm getting a compile failure from a change contained in the recent commit 9462db5

```
mfuns.c: In function ‘mfn_flags’:
mfuns.c:3546:15: error: implicit declaration of function ‘flag_object’ [-Wimplicit-function-declaration]
 3546 |     buf3[0] = flag_object(obj);
      |               ^~~~~~~~~~~
make[1]: *** [Makefile:90: mfuns.o] Error 1
```

Not much more to add!

wyld-sw:
These changes:
- remove unused return values and prototypes in prochelp.c and resolver.c
- address a time_t inconsistency in timequeue.c
- make the callback function inline in p_mcp.c
- add "muf:7.0" (arbitarilty the major server version) as a #define in mcppkgs.c
- return an unsupported message for the "user" category for simpleedit in mcppkgs.c
- rename add_ and delete_player in player.c
- fix uninitialized variables in sanity.c
- write to status log during emergency signal in fbsignal.c
- ~remove dead stores in mufevent.c~
- combine timestamp storage with host_add* functions in resolver.c 
- fix compiler warnings relating to string truncation and a missing prototype

### DIFF
--- a/src/fbtime.c
+++ b/src/fbtime.c
@@ -317,17 +317,18 @@ timestr_full(long dtime)
 const char *
 timestr_long(long dtime)
 {
-    static char buf[SMALL_BUFFER_LEN], value[SMALL_BUFFER_LEN];
+    static char buf[SMALL_BUFFER_LEN];
     const char *str[7] = { "year", "month", "week", "day", "hour", "minute", "second" };
     int div[7] = { 31556736, 2621376, 604800, 86400, 3600, 60, 1 };
 
-    *buf = *value = 0;
+    *buf = 0;
 
     for (int i = 0; i < 7; i++) {
         if (dtime >= div[i]) {
             long temp = dtime / div[i];
-            snprintf(value, sizeof(value), "%s%s%ld %s%s", buf, (*buf ? ", " : ""), temp, str[i], temp != 1 ? "s" : "");
-            strncpy(buf, value, sizeof(buf));
+            size_t used = strlen(buf);
+            snprintf(buf + used, sizeof(buf) - used, "%s%ld %s%s",
+                     used ? ", " : "", temp, str[i], temp != 1 ? "s" : "");
             dtime %= div[i];
         }
     }
--- a/src/fbtime.c
+++ b/src/fbtime.c
@@ -320,15 +320,21 @@ timestr_long(long dtime)
     static char buf[SMALL_BUFFER_LEN];
     const char *str[7] = { "year", "month", "week", "day", "hour", "minute", "second" };
     int div[7] = { 31556736, 2621376, 604800, 86400, 3600, 60, 1 };
+    size_t used = 0;
 
     *buf = 0;
 
     for (int i = 0; i < 7; i++) {
         if (dtime >= div[i]) {
             long temp = dtime / div[i];
-            size_t used = strlen(buf);
-            snprintf(buf + used, sizeof(buf) - used, "%s%ld %s%s",
-                     used ? ", " : "", temp, str[i], temp != 1 ? "s" : "");
+            int printed = snprintf(buf + used, sizeof(buf) - used, "%s%ld %s%s",
+                                   (used ? ", " : ""), temp, str[i], temp != 1 ? "s" : "");
+            if (printed > 0) {
+                used += printed;
+                if (used >= sizeof(buf)) {
+                    used = sizeof(buf) - 1;
+                }
+            }
             dtime %= div[i];
         }
     }
--- a/src/mfuns.c
+++ b/src/mfuns.c
@@ -3545,7 +3545,7 @@ mfn_flags(MFUNARGS)
     buf3[1] = '\0';
     buf3[0] = flag_object(obj);
     flag_list(obj, buf2, sizeof(buf2));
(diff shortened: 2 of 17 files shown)


src/fbtime.c
```diff
@@ -317,17 +317,18 @@ timestr_full(long dtime)
...
-    *buf = *value = 0;
+    *buf = 0;
 
     for (int i = 0; i < 7; i++) {
         if (dtime >= div[i]) {
             long temp = dtime / div[i];
-            snprintf(value, sizeof(value), "%s%s%ld %s%s", buf, (*buf ? ", " : ""), temp, str[i], temp != 1 ? "s" : "");
-            strncpy(buf, value, sizeof(buf));
+            size_t used = strlen(buf);
+            snprintf(buf + used, sizeof(buf) - used, "%s%ld %s%s",
```
PointsDragon: snprintf returns the number of bytes printed - so rather than doing O(n) via strlen on each iteration pass, you can keep this O(1) by simply setting used to 0 in the function, then incrementing used by the return value on each loop pass.

src/mfuns.c
```diff
@@ -3545,7 +3545,7 @@ mfn_flags(MFUNARGS)
     buf3[1] = '\0';
     buf3[0] = flag_object(obj);
     flag_list(obj, buf2, sizeof(buf2));
-    snprintf(buf, sizeof(buf), "%s%s", buf3, buf2);
+    snprintf(buf, buflen, "%s%s", buf3, buf2);
```
PointsDragon, line 3548: Good catch

src/mufevent.c
```diff
@@ -81,52 +81,20 @@ muf_event_process_free(struct mufevent_process *ptr)
...
-    ptr->next = NULL;
-    ptr->player = NOTHING;
-
     if (ptr->fr) {
         if (PROGRAM_INSTANCES(ptr->prog)) {
             prog_clean(ptr->fr);
         }
-
-        /* @TODO: I know it's nitpicky, but this continues to be silly */
-        ptr->fr = NULL;
```
PointsDragon, line 101: Note the comment to this function - this may require a bit more digging, or the function commentary needs to be updated.

src/p_array.c
```diff
@@ -1692,7 +1692,12 @@ prim_array_get_propdirs(PRIM_PROTOTYPE)
     propadr = first_prop(ref, dir, &pptr, propname, sizeof(propname));
 
     while (propadr) {
-        snprintf(buf, sizeof(buf), "%s%c%s", dir, PROPDIR_DELIMITER, propname);
+        int used = snprintf(buf, sizeof(buf), "%s%c%s", dir, PROPDIR_DELIMITER, propname);
+
+        if (used < 0 || (size_t) used >= sizeof(buf)) {
```
PointsDragon, line 1697: The spec for snprintf for truncation return values is a wee bit unclear - but I believe you may want this to be an explicitly greater than, as opposed to greater or equal to.  May be worth running a quick test here.  Older versions of glib return -1 for truncation conditions.

src/p_array.c
```diff
@@ -2179,6 +2190,10 @@ prim_array_put_propvals(PRIM_PROTOTYPE)
                     *propname = '\0';
             }
 
+            if (used < 0 || (size_t) used >= sizeof(propname)) {
```
PointsDragon, line 2193: Per comment above.

src/p_mcp.c
PointsDragon: You could just cast this within the free and avoid the use of the local variable.

src/sanity.c
```diff
@@ -1132,11 +1132,19 @@ create_lostandfound(dbref * player, dbref * room)
     PUSH(*room, CONTENTS(GLOBAL_ENVIRONMENT));
     SanFixed(*room, "Using %s to resolve unknown location");
 
-    while (lookup_player(player_name) != NOTHING && strlen(player_name) < (unsigned int)tp_player_name_limit) {
-        snprintf(player_name, tp_player_name_limit+1, "lost+found%d", ++temp);
+    player_name_len = snprintf(player_name, tp_player_name_limit+1,
+                               "lost+found");
+
+    for (temp = 1;
```
PointsDragon: If we're not using 'temp' outside of the for loop, we can declare 'int temp = 1' here in the loop init and remove it from the function scope above.

src/sanity.c
```diff
@@ -1315,18 +1323,20 @@ find_misplaced_objects(void)
 
                 case TYPE_PLAYER: {
                     char *name = malloc(tp_player_name_limit+1);
+                    int temp;
 
-                    /* Loop to find a name we can use */
-
-                    /* @TODO: turn this into a for loop */
-                    int temp = 0;
+                    for (temp = 1; ; temp++) {
```
PointsDragon: Per above comment.

PointsDragon (commented):
Solid work, couple of minor nits for your consideration.

PointsDragon (approved):

PR #780, opened 2026-07-11
welcome screen processing refactor

wyld-sw:
This attempts a refactor of welcome screen logic, based on the related TODO.

Added sysparms:
- `log_hosts`: if no, logs hosts as "hidden host". Defaults to yes. MUF and Wizards can currently still view the host.
- `wizonly_bootmesg`: the string sent to the descriptor if the server is in wizonly mode.

Added functions:
- `boot_restricted`: sends the "reason" message and sets the descriptor to be booted.
- `tokenize_as`: I like the use of the previous `check_connect`, and this could be useful in other contexts.

Changed log lines. I could see these being a minor - and temporary - inconvenience for log parsers.
- Uses `REGISTRATION REQUIRED` instead of `FAILED CREATE` if registration=yes.
- Combines SSL cipher name display with creation/connection message.
- Adds the colon to `CREATED:` on successful player creation.

I think I've tested each path, and did not notice any logic deviation. However, please feel free to switch into this PR branch and test!

### DIFF
--- a/src/fbstrings.c
+++ b/src/fbstrings.c
@@ -2022,3 +2022,33 @@ toupper_string(char **s)
    for (char *p = *s; *p; p++)
         *p = toupper(*p);
 }
+
+/**
+ * Tokenizes a string into an array of fixed-size pre-allocated buffers.
+ *
+ * @param msg             The input string.
+ * @param buffer_array    Array of pointers to your destination buffers.
+ * @param array_size      How many buffers you passed.
+ * @param buffer_capacity The maximum capacity of EACH individual buffer.
+ */
+void
+tokenize_to_array(const char *msg, char **buffer_array, int array_size, size_t buffer_capacity)
+{
+    if (!msg || !buffer_array || array_size <= 0 || buffer_capacity == 0) return;
+    
+    const char *current = msg;
+    const char *delimiters = " \t";
+
+    for (int i = 0; i < array_size; i++) {
+        buffer_array[i][0] = '\0';
+        current += strspn(current, delimiters);
+        if (*current == '\0') break;
+
+        size_t token_len = strcspn(current, delimiters);
+        size_t copy_len = (token_len >= buffer_capacity) ? (buffer_capacity - 1) : token_len;
+
+        memcpy(buffer_array[i], current, copy_len);
+        buffer_array[i][copy_len] = '\0';
+        current += token_len;
+    }
+}
--- a/src/interface.c
+++ b/src/interface.c
@@ -1746,7 +1746,7 @@ process_welcome_input(struct descriptor_data *d, const char *msg)
     if (do_connect) {
         log_status("CONNECTED: %s(%d), %s", NAME(player), player, connect_string);
     } else {
-        log_status("CREATED %s(%d), %s", NAME(player), player, connect_string);
+        log_status("CREATED: %s(%d), %s", NAME(player), player, connect_string);
     }
 
     d->connected = 1;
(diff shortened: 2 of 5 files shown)


wyld-sw:
@tanabi or @PointsDragon - how does this look? Does it need adjustment before merge?

tanabi (approved):

src/fbstrings.c
```diff
@@ -2022,3 +2022,33 @@ toupper_string(char **s)
...
+{
+    if (!msg || !buffer_array || array_size <= 0 || buffer_capacity == 0) return;
+    
+    const char *current = msg;
+    const char *delimiters = " \t";
+
+    for (int i = 0; i < array_size; i++) {
+        buffer_array[i][0] = '\0';
+        current += strspn(current, delimiters);
+        if (*current == '\0') break;
```
PointsDragon, line 2045: You likely want to make this a continue, not a break, so that all members of the buffer_array get NULL terminated before the function returns.

src/interface.c
```diff
@@ -1645,154 +1669,103 @@ connect_player(const char *name, const char *password)
...
-#else
-         "",
+    if (d->ssl_session) {
+        snprintf(connect_string, sizeof(connect_string), "descriptor %d, securely (via %s) from %s",
+                 d->descriptor, SSL_get_cipher_name(d->ssl_session), host_log);
+    } else {
 #endif
-    d->hostname);
-
-    parse_connect(msg, command, user, password);
```
PointsDragon, line 1665: I don't see parse_connect() being removed from interface.c, though I don't think anything is actually using it any longer after this change?

src/interface.c
```diff
@@ -1645,154 +1669,103 @@ connect_player(const char *name, const char *password)
...
-
-                    /* cks: someone has to initialize this somewhere. */
-                    PLAYER_SET_BLOCK(d->player, 0);
-                    spit_file_segment(player, tp_file_motd, "");
-                    announce_connect(d->descriptor, player);
-                    con_players_curr++;
-                }
-            }
-        } else {
+    if (!strncmp(command, "help", 4)) {
```
PointsDragon, line 1699: Per my commentary else-where, we can get rid of repeated strncmp by converting the first two chars of command into a radix value, then replacing this if/else tree into an integer switch statement.

src/interface.c
```diff
@@ -1645,154 +1669,103 @@ connect_player(const char *name, const char *password)
...
+        player = create_player(user, password, error);
+        if (player == NOTHING) {
+            queue_ansi(d, error);
+            queue_write(d, "\r\n", 2);
+            log_status("FAILED CREATE: '%s', %s", user, connect_string);
+            return;
+        }
+    }
+
+    if (do_connect) {
```
PointsDragon, line 1746: Since the two log_status lines are the same except the pre-pended action, you could use your selection tree above to also populate an 'action' string variable, then have a simple log_status line below that is "%s: %s(%d), %s" with no additional 'if' required.

src/interface.c
```diff
@@ -1645,154 +1669,103 @@ connect_player(const char *name, const char *password)
...
+        do_create = 1;
     } else if (!*command) {
-        /* do nothing */
+        return;
     } else {
         welcome_user(d);
+        return;
     }
+
+    if (do_connect) {
```
PointsDragon, line 1723: Since we can't be both 'do_connect' and 'do_create', we could replace these with a state variable instead and again do a simple switch here.  If we want to maintain readability, use an enum or define constants to make it text but keep it numeric behind the scenes.

PointsDragon (commented):
Should work as committed, but you might want to take a peek at the comments and see if they're worth acting on.

PR #787, opened 2026-07-26
Removes divide-by-zero flaw, addresses non-transitivity

### LINKED ISSUE #179: array_tree_compare is not transitive, sometimes divides by zero
array_tree_compare tries to compare floats using a margin of DBL_EPSILON, but this means that its comparison is not transitive. It also uses division by one of the values being compared to scale this comparison, which sometimes divides by zero.

The division by zero definitely needs to get fixed. Any change to array_tree_compare probably changes the user-visible semantics of MUF dictionaries.

wyld-sw:
This restates the division as a multiplication to avoid the NaN resulting from division by zero. 0 and 0.0 should no longer be smaller than each other.

It also extracts the comparison routine and uses a switch instead of multiple ifs.

Resolves #179.

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -368,6 +368,32 @@ array_tree_compare(const array_iter * a, const array_iter * b, int case_sens) {
 }
 
 /**
+ * Compares two double-precision floating-point values using Knuth's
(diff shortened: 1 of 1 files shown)


PointsDragon (approved):
LGTM

PR #798, opened 2026-09-03
Fix page to work with connection prims gone

tanabi:
I finally got around to upgrading my MUCK and quickly discovered page wasn't working right. A super simple fix here.


PointsDragon (approved):
LGTM

